PR #83, opened 2014-06-09
add test showing how hooks fail if not run last

neonstalwart:
hooks should be run after other properties have been applied so that their changes are not undone


Matt-Esch:
Convinced this is not a thing that needs to land, but feel free to reopen if you still think this is a problem.

PR #138, opened 2014-12-08
passing children as arguments

heavyk:
coming from hyperscript original I was sad to see varargs not passed as children. if you'd like to add it, it's a few lines..

cool project dude

### DIFF
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -83,7 +83,7 @@ function h(tagName, properties, children) {
 
     if (children !== undefined && children !== null) {
         addChild(children, childNodes, tag, props)
-        for (var i = 0,
+        if (arguments.length > 3) for (var i = 0,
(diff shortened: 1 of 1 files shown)


killercup:
FYI @Raynos, @Matt-Esch: This is currently preventing me from using Babel's JSX transform with virtual-dom. (I totally understand your reasoning, though, and consider the array argument to be a nicer API.)

Raynos:
@killercup https://github.com/Raynos/mercury/tree/master/examples/unidirectional/jsx

I've got a working example with jsx. See if that works for you.

killercup:
Thanks. I've got it working by using jsx-transform and babel together. It's
not as nice as I had hoped it could be (it's transpiling ever file twice),
but it does the job.
Jake Verbaten <notifications@github.com> schrieb am So., 19. Juli 2015 um
02:15:

PR #154, opened 2014-12-26
Accept number as a vText node

jtangelder:


unframework:
Since the intent seems to be to emulate Hyperscript as a DSL, it could be useful to mimic the library's autoconversion behaviour for primitives.

Here is the vanilla Hyperscript implementation: https://github.com/dominictarr/hyperscript/blob/master/index.js#L37

I would definitely vote in favour of keeping the two implementations consistent. This way legacy Hyperscript code can be ported cleanly to use virtual-hyperscript implementation.

unframework:
Just for the record, I agree that implicit type coercion is bad, but this could still keep underlying implementation (VDom) tight, but making the glue (Hyperscript DSL) loose.

kuraga:
@unframework sorry what do you propose, exactly?

unframework:
I was just presenting a counter-argument to the "implicit type coercion is the devil" reply.

alexmingoia:
I disagree this should be merged. Implicit type conversion should not happen at this layer. If you want to create a text node with numbers, simply pass it as a string before it gets there.

substack:
Type coercion here would be great. The alternative is to blow up with a very difficult to diagnose error message for something that shouldn't matter.

Raynos:
This has now been accepted and landed. 

There was a lot of ask for this feature so we've added it.

Issue #104, opened 2014-08-27
Question about controlling position at which to insert element.


ih2502mk:
Hi,

I'm using mercury and I have an app that looks a lot like todo list i.e. a form to add new element and a list of added elements. However in my case I want to have a transition when new element is added and I want a new element to be added to the top of the list "layoutwise".
I could figure out how to use hooks to achieve transition effect - that worked perfect for me.

The problem is with the way DOM is patched when new item is added to an array of items in the app state. Transition always happens on the last item in the list.

To give you an idea, I have:
```javascript
state.items = mercury.array([]);
```
and when I "submit" a form: 
```javascript
state.items.unshift(Item( /* data that came from the form*/ ));
```
and rendering looks like this:
```javascript
function renderItemsList(state) {  
  return h('ul#item-list', [
    state.items.map(function(item) {return Item.render(item);})
  ]);
}
``` 

My assumption is that a process of creating a patch looks something like this:
```
original Vtree:
ul
 - item3
 - item2
 - item1

updated Vtree
ul
 - item4
 - item3
 - item2
 - item1

and the patch looks like this:
 - change item3 to item4
 - change item2 to item3
 - change item1 to item2
 - insert item1
```

Whereas I would like to have a patch that looks like: 
```
 - insert item4 before item3
```

I'd like to know if my assumption of how patch is created is correct to begin with. And also if there is a way to achieve the desired outcome.

Here is a requirebin of an application that I was talking about: http://requirebin.com/?gist=086fb96724577864a3b6

AlexGalays:
Not sure, I don't know the codebase yet. It looks like you already put in a few optimizations? How does that reverse index work?

I don't really get why a new node inserted at index 3, is actually appended (INSERT patch) then moved. I guess that made other uses cases simpler?

Would a naive optimization approach work?
```
if (num(INSERT patches) == n-1) { // n = total number of patches; bail fast in most occasion
    check the MOVE patch; for each of these (new) n-1 indices, if they're being moved to a consecutive 0 index, delete all the other scheduled moves.
}
```
use-case-specific optimisations like this one may feel weird but we have to remember this is being used in UIs where prepend-the-newest-items is a very common operation (e.g Facebook top right activity panel)

For info, knockout computes the levenshtein distance between the 2 arrays to compute the least amount of patches, I don't think you can beat that at the DOM level but it's unclear whether the JS overhead makes it worth it.

AlexGalays:
[...]

PR #156, opened 2015-01-02
optimize the case where a new element is added to front of list

neonstalwart:
@AlexGalays this PR provides an optimization for the specific case you mentioned in https://github.com/Matt-Esch/virtual-dom/issues/104#issuecomment-68447534

the patches will still look the same but i've optimized `reorderChildren` to better deal with node reordering.  previously the algorithm only recognized when the node at the current insert point was going to move towards the end of the list of children and inserted after it but now i've changed it to realize when inserting a node before the current insert point will put that node in the right place and hence reduce the number of moves needed.

in your case it was previously doing these 2 patches:
  1. add the `'3'` node to the end of the list of children
  1. reorder the children
    * the `'3'` node needs to move to position 0
    * the element at position 0 (the current insert position) is scheduled to move towards the end of the list of children so increase our insertion offset and insert `'3'` after this node and then when the node at 0 moves, `'3'` will be at position 0
    * move the insert point 
    * the  `'1'` node needs to move to the current insert point
    * the node currently at the insert point is scheduled to move towards the end of the list of children so increase the insertion offset and insert `'1'` after this node
    * move the insert point
    * the `'2'` node needs to move to the current insert point
    * move the `'2'` node to the current insert point

this led to each node needing to be moved to get them into the right position.  now the 2 patches are handled like this:
  1. add the `'3'` node to the end of the list of children
  1. reorder the children
    * the `'3'` node needs to move to position 0
    * the element at position 0 is scheduled to move towards the end of the list but if we insert this node before it, it will be in the right place so insert `'3'` before the current insert point.
    * move the insert point
    * the `'1'` node needs to move to the current insert point
    * the node at the current insert point is already `'1'` so do nothing
    * move the insert point
    * the `'2'` node needs to move to the current insert point
    * the node at the current insert point is already `'2'` so do nothing

### DIFF
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -154,9 +154,8 @@ function reorderChildren(domNode, bIndex) {
             insertNode = childNodes[i + insertOffset] || null
             insertedLength = 0
             while (node !== insertNode && insertedLength++ < chainLength) {
-                nextSibling = node.nextSibling;
                 domNode.insertBefore(node, insertNode);
-                node = nextSibling;
+                node = children[move + insertedLength];
             }
 
             // the moved element came from the front of the array so reduce the insert offset
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -133,22 +133,34 @@ function reorderChildren(domNode, bIndex) {
     var move
     var node
     var insertNode
-    for (i = 0; i < len; i++) {
+    var chainLength
+    var insertedLength
+    var nextSibling
+    for (i = 0; i < len;) {
         move = bIndex[i]
+        chainLength = 1
         if (move !== undefined && move !== i) {
+            // try to bring forward as long of a chain as possible
+            while (bIndex[i + chainLength] === move + chainLength) {
+                chainLength++;
+            }
+
             // the element currently at this index will be moved later so increase the insert offset
-            if (reverseIndex[i] > i + 1) {
+            if (reverseIndex[i] > i + chainLength) {
                 insertOffset++
             }
 
             node = children[move]
             insertNode = childNodes[i + insertOffset] || null
-            if (node !== insertNode) {
-                domNode.insertBefore(node, insertNode)
+            insertedLength = 0
+            while (node !== insertNode && insertedLength++ < chainLength) {
+                nextSibling = node.nextSibling;
+                domNode.insertBefore(node, insertNode);
+                node = nextSibling;
             }
 
             // the moved element came from the front of the array so reduce the insert offset
-            if (move < i - 1) {
+            if (move + chainLength < i) {
                 insertOffset--
             }
         }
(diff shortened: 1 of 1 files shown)


AlexGalays:
Thanks in helping getting this fixed;
Unfortunately, it's still broken in my app, after applying your changes to patch-op locally.

Should we expose `reorderChildren` so we can write a test for this specific scenario? Or fix it at the patch level, although it's probably much more work.

neonstalwart:
@AlexGalays how are you determining that your app is broken?  can you make a gist that demonstrates the same problem you see in your app.  this patch reduces the number of times the DOM is touched so if something else is not working i want to know how to tell that it's broken so i can try to fix that issue.

AlexGalays:
I will produce a gist soon.
Your patch indeed fixes the case where only one new item is added at index 0. But it doesn't work if N are added.

AlexGalays:
Here's a gist you can use to reproduce the issue:  https://gist.github.com/AlexGalays/9d57450aef5df324cf93

Just paste it in your requirebin. When clicking on 'add', only the **new** items should flash, not all.

neonstalwart:
@AlexGalays try my latest changes.  the rationale with these changes is that new nodes are always added to the end so we should try and bring as many of those nodes forward at once as possible.

neonstalwart:
fyi @Raynos and @Matt-Esch these tests failed due to min-document not implementing `node.nextSibling` but i re-implemented the patch without `nextSibling`

AlexGalays:
@neonstalwart  Works nicely, thank you :+1:

AlexGalays:
@Raynos, @Matt-Esch, @neonstalwart 

Only discovered this now, but this change breaks the patching of a list of key'ed elements. I get random behaviors such as of-by-one errors (the sibling was removed instead of the correct node) and duplicated, lingering nodes at the end of the ul. I've seen applyPatch -> removeNode being called with the wrong element/vnode combination. 
In 1.1.0+, It works fine if the elements are not key'ed. It works fine in 1.0.0 even if the elements are key'ed.

We really want to keep @neonstalwart's work because CSS transitions/animations on various browsers are capricious enough without adding another level of entropy with elements being shifted around for no reason.

I will try to produce a requirebin showing the issue later today.

neonstalwart:
for my own reference... http://jsfiddle.net/neonstalwart/dg66onon/ since jsfiddle is easier to debug.

i'll look into this next chance i get.

vdom/patch-op.js
```diff
@@ -133,22 +133,33 @@ function reorderChildren(domNode, bIndex) {
...
+        chainLength = 1
         if (move !== undefined && move !== i) {
+            // try to bring forward as long of a chain as possible
+            while (bIndex[i + chainLength] === move + chainLength) {
+                chainLength++;
+            }
+
             // the element currently at this index will be moved later so increase the insert offset
-            if (reverseIndex[i] > i) {
+            if (reverseIndex[i] > i + chainLength) {
```
neonstalwart: @AlexGalays if you've got a chance before i get around to work on tests and a PR, could you try with this being `>=` instead of `>` - i suspect that may fix it but i won't know for sure until i can sit down and right tests.

PR #158, opened 2015-01-05
additional svg attributes added

sunwukung:
stroke attributes added to help with rendering svg line elements.
also added style attribute for a bit of flexibility


Matt-Esch:
stroke is fixed in #163 - style is **not** an attribute. It's a DOM property, which has a object value such as:

```style: { border: "1px solid #000" }```

It shouldn't be added as an attribute.

sunwukung:
i was dubious myself. i see you've added the legit attributes, many thanks

Issue #137, opened 2014-12-08
Using virtual-dom without browserify


pm-debjit-biswas:
I am currently building a standalone file (with global variable) using browserify separately and including it in my project. Is there a better way to do this, because the virtual-dom dependencies are not listed in the project package.json file?

In future I would also like to use other libraries associated with virtual-dom such as vdom-thunk.

mato75:
npm run dist throws errors:


virtual-dom@1.1.0 dist /Users/matjaz/Dropbox/GitHub/virtual-dom
browserify --standalone virtual-dom index.js > dist/virtual-dom.js

sh: dist/virtual-dom.js: No such file or directory

npm ERR! virtual-dom@1.1.0 dist: `browserify --standalone virtual-dom index.js > dist/virtual-dom.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the virtual-dom@1.1.0 dist script.
npm ERR! This is most likely a problem with the virtual-dom package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     browserify --standalone virtual-dom index.js > dist/virtual-dom.js
npm ERR! You can get their info via:
npm ERR!     npm owner ls virtual-dom
npm ERR! There is likely additional logging output above.
npm ERR! System Darwin 14.0.0
npm ERR! command "/opt/local/bin/node" "/opt/local/bin/npm" "run" "dist"
npm ERR! cwd /Users/matjaz/Dropbox/GitHub/virtual-dom
npm ERR! node -v v0.10.26
npm ERR! npm -v 1.4.4
npm ERR! code ELIFECYCLE
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /Users/matjaz/Dropbox/GitHub/virtual-dom/npm-debug.log
npm ERR! not ok code 0

ericgj:
`mkdir -p dist` ?

mtyaka:
@mato75 That scripts assumes browserify is installed globally on your system: `npm install -g browserify`

mato75:
thx both of you :). It worked.

I still agree with ericgj, that this should be already included in every release.

neonstalwart:
cleaning up old tickets - open another issue if there's something here which is still not resolved

PR #168, opened 2015-01-13
allow `npm run dist` to work after npm install

bitinn:
* added `dist` folder so that browserify don't complain about missing directory, exclude build for now.
* added browserify to devDependencies, people with global browserify might not like it, but i think `npm install --production` is better than requiring `npm install browserify -g`

avoid confusion like #137


Raynos:
lgtm.

PR #171, opened 2015-01-15
isHook() returns true for unhook type hooks

Zolmeister:


Raynos:
interesting, I've never considered just unhook on its own.

What's a use case for it ?

Zolmeister:
The original use was to trigger a side-effect upon unloading a component (persisting state).
A more correct use case has been binding/unbinding global events without using an event delegator

e.g.

```coffee
class zoriumComponent
  constructor: ->
   @listener = (e) -> console.log 'scroll event'
    window.addEventListener 'scroll', @listener
  
  # unhook
  onBeforeUnmount: =>
    window.removeEventListener @listener
```

Raynos:
lgtm.

neonstalwart:
@Zolmeister i'm curious why you don't use `hook` to add the listener?

Zolmeister:
@neonstalwart  I usually reserve hook as a method for grabbing the DOM node, so structurally it didn't make sense. Also, [Zorium](https://github.com/claydotio/zorium) internally separates the two hooks and this issue caused our tests to fail when testing the hooks separately.

PR #172, opened 2015-01-16
Fix AttributeHook and use it for xlink SVG attributes

mtyaka:
Xlink attributes have to be set with `setAttributeNS` into the xlink namespace, otherwise they don't have any effect.
Here's a demonstration with two `<image>` elements - only the one that has its `xlink:href` attribute set with `setAttributeNS` shows up: http://jsfiddle.net/3phtrLLt/

The `AttributeHook` test depends on https://github.com/Raynos/min-document/pull/18.


Raynos:
lgtm.

virtual-hyperscript/hooks/attribute-hook.js
```diff
@@ -4,7 +4,7 @@ module.exports = AttributeHook;
 
 function AttributeHook(namespace, value) {
     if (!(this instanceof AttributeHook)) {
-        return new AttributeHook(value);
+        return new AttributeHook(namespace, value);
```
Matt-Esch: :+1:

virtual-hyperscript/svg.js
```diff
@@ -42,6 +44,11 @@ function svg(tagName, properties, children) {
             continue;
         }
 
+        if (key.substr(0, 6) === 'xlink:') {
```
Matt-Esch: I think we might be able to make this more efficient by putting the namespace information in the attribute lookup table

mtyaka: @Matt-Esch Thanks for checking it out. I pushed a change to move the namespace into the attribute lookup table and included the `xml:` and `ev:` namespaces as well. Can you take another look, please?

virtual-hyperscript/hooks/attribute-hook.js
neonstalwart: semver question... is this a breaking change to a public API?  specifically, it's obviously a breaking change but is it a public API?

PR #173, opened 2015-01-19
Clarify diff props & attributes tests

jesseditson:
I was reading through the tests & diff.js, and it wasn't immediately clear to me what things were doing, so I split diffProps into it's own module and expanded an attributes test.

Not sure what the style guide on this repo is, but figured having 1 more test and making things more modular might be appreciated.


Raynos:
Tests are always welcome :)

PR #175, opened 2015-01-20
Don't try to hook unhook-only hooks.

mtyaka:
`isHook()` returns true for hooks with only the `unhook` method
as of 75e1f61.

Don't call `hook()` on hooks that don't have a `hook` method.

### DIFF
--- a/test/hook.js
+++ b/test/hook.js
@@ -263,6 +263,30 @@ test("property-replacing diff calls unhook", function (assert) {
   assert.end()
 })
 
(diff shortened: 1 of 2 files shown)


Raynos:
lgtm. We might want an `isHook()` and `isUnhook()` method. I'll let @Matt-Esch decide.

Matt-Esch:
I like the idea of splitting this out into `isHook` and `isUnhook`, but I'll create a separate task for it.

PR #177, opened 2015-01-22
Pass next value to the unhook() method and use it to fix a bug in AttributeHook

mtyaka:
The patch adds a third parameter to `unhook()` - the next instance of the hook, or `undefined` if the hook is removed.

This is consistent with `hook()`, which receives the previous instance as the third parameter.

As of 4aedb4a3697b7e3f3eee11f769e20a02de5c534f, `unhook()` is invoked every time the hook gets patched. The next instance can be used to determine whether `unhook()` can avoid doing unnecessary work, for example:

    MyHook.prototype.unhook = function(node, prop, next) {
      if (next && next.value === this.value) {
        return
      }
      // ... do work ...
    }

The new parameter is then used to fix a bug in `AttributeHook` where attribute that didn't change were accidentally removed when re-rendering the node.

### DIFF
--- a/virtual-hyperscript/hooks/attribute-hook.js
+++ b/virtual-hyperscript/hooks/attribute-hook.js
@@ -19,8 +19,12 @@ AttributeHook.prototype.hook = function (node, prop, prev) {
     node.setAttributeNS(this.namespace, prop, this.value);
 };
 
-AttributeHook.prototype.unhook = function (node, prop) {
-    var colonPosition = prop.indexOf(':')
-    var localName = colonPosition > -1 ? prop.substr(colonPosition + 1) : prop
+AttributeHook.prototype.unhook = function (node, prop, next) {
+    if (next !== undefined) {
+        return;
+    }
+
+    var colonPosition = prop.indexOf(':');
+    var localName = colonPosition > -1 ? prop.substr(colonPosition + 1) : prop;
     node.removeAttributeNS(this.namespace, localName);
 };
--- a/virtual-hyperscript/hooks/attribute-hook.js
+++ b/virtual-hyperscript/hooks/attribute-hook.js
@@ -12,7 +12,9 @@ function AttributeHook(namespace, value) {
 }
 
 AttributeHook.prototype.hook = function (node, prop, prev) {
-    if (prev && prev.value === this.value) {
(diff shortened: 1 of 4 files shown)


neonstalwart:
if there's a bug in AttributeHook we should fix it.  do you have a test case that demonstrates that bug?

when a node gets re-used, i'm not sure it's safe to assume that the next hook is the same kind of hook that is being unhooked.  the hook being unhooked might be an AttributeHook and the new hook might be of another kind.  you really want the AttributeHook to unhook in this case even though there's a new hook.  is there a flaw in my logic?

mtyaka:
@neonstalwart Thanks a lot for your insight. The expanded AttributeHook test demonstrates the problem I was trying to fix - this line fails without the patch: https://github.com/mtyaka/virtual-dom/commit/35a0a43d35f9bac4bc1d51170954319c0b37b96b#diff-e11e1549819cd254dcc361c130e6d85cR28

I think your logic is correct, I didn't really consider the scenario where a hook gets replaced with another type of hook. I added https://github.com/mtyaka/virtual-dom/commit/7ae386c813964321b25107fbb4e0bb1d8fa812f2 which I believe fixes the issue, please check it out.

virtual-hyperscript/hooks/attribute-hook.js
```diff
@@ -12,15 +12,21 @@ function AttributeHook(namespace, value) {
 }
 
 AttributeHook.prototype.hook = function (node, prop, prev) {
-    if (prev && prev.value === this.value) {
+    if (prev instanceof AttributeHook &&
```
Matt-Esch: We aggressively avoid `instanceof` in virtual-dom to prevent errors across boundaries and duplication. Favor putting a `type` field on the attribute hook prototype and compare against that.

Matt-Esch:
I think this would be good for preventing any unwanted heavy side effects. Hooks are only every equal by reference, which has helped us avoid a huge number of unwanted hook/unhook calls, but I'm sure there is room for improvement.

We do need to get rid of the instanceof checks before this can be landed.

mtyaka:
@Matt-Esch Thanks, I replaced the `instanceof` check with a `type` property check and rebased against master.

PR #178, opened 2015-01-28
Update VNode and VText package names

teropa:
VNode is in "virtual-dom/vnode/vnode", not "virtual-dom/vtree/vnode". Ditto for VText.


Raynos:
lgtm.

PR #180, opened 2015-02-05
improved performance of parseTag method for the common case

melnikov-s:
Hello,

I've been playing around with virtual-dom lately and created my own todomvc app with it. When testing I noticed that my version was significantly slower than mercurys reference example. This seemed odd as they both used the virtual-dom library. Upon doing some profiling I found the culprit to be the browser-split method which was introduced in 900d332 , I'm guessing in order to deal with some browser inconsistencies with splitting a string using regex.

Anyways, in this pull request I've avoided using the browser-split method if the tag passed into `parseTag` does not contain any id's or class selectors. This greatly increased the performance of the common case (a plain tag) but heavy usage of `div#id.selector` notation will remain slow. 

Was browser-split introduced because of an existing incompatibility of supported browsers or as a pre-emptive measure ? Can it be changed back to `String.prototype.split` ?

Here are the basic benchmarks:

``` javascript
       var renderAll = function() {
            var list = [];
            for (var i =0; i < 1000; i++) list.push(renderOne(String(i));

            return h('div', list);
        }

        var renderOne = function(v) {
            return h('div', [
                h('input', {type: 'text'}, v ),
                h('div', v),
                h('label', v),
                h('span', v),
                h('p', v)
            ]);
        }
```

Before:

``` javascript
var t =performance.now() ; renderAll(); performance.now() - t; //287.14299999410287
```

After:

``` javascript
var t =performance.now() ; renderAll(); performance.now() - t; //20.79000003868714
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "email": "matt@mattesch.info"
   },
   "dependencies": {
-    "browser-split": "0.0.1",
     "error": "^4.3.0",
     "ev-store": "^7.0.0",
     "global": "^4.3.0",
--- a/virtual-hyperscript/parse-tag.js
+++ b/virtual-hyperscript/parse-tag.js
@@ -1,10 +1,5 @@
 'use strict';
 
-var split = require('browser-split');
-
-var classIdSplit = /([\.#]?[a-zA-Z0-9_:-]+)/;
-var notClassId = /^\.|#/;
-
 module.exports = parseTag;
 
 function parseTag(tag, props) {
@@ -14,16 +9,13 @@ function parseTag(tag, props) {
 
     var noId = !(props.hasOwnProperty('id'));
 
-    var tagParts = split(tag, classIdSplit);
-    var tagName = null;
+    var tagParts = splitTag(tag);
 
-    if (notClassId.test(tagParts[1])) {
-        tagName = 'DIV';
-    }
+    var tagName = tagParts[0] || 'DIV';
 
     var classes, part, type, i;
 
-    for (i = 0; i < tagParts.length; i++) {
+    for (i = 1; i < tagParts.length; i++) {
         part = tagParts[i];
 
         if (!part) {
@@ -32,9 +24,7 @@ function parseTag(tag, props) {
 
         type = part.charAt(0);
 
-        if (!tagName) {
-            tagName = part;
-        } else if (type === '.') {
+        if (type === '.') {
             classes = classes || [];
             classes.push(part.substring(1, part.length));
         } else if (type === '#' && noId) {
@@ -51,4 +41,32 @@ function parseTag(tag, props) {
     }
 
     return props.namespace ? tagName : tagName.toUpperCase();
+}
+
+
+function splitTag(tag) {
+
+    var classIndex, idIndex,
+        remaining = tag,
+        parts = [],
+        last = '';
+
+    do {
+        idIndex = remaining.indexOf('#');
+        classIndex = remaining.indexOf('.');
+        if ((idIndex === -1 || idIndex > classIndex) && classIndex !== -1) {
+            parts.push(last + remaining.substr(0, classIndex));
+            last = '.';
+            remaining = remaining.substr(classIndex + 1);
+        } else if (idIndex !== -1){
+            parts.push(last + remaining.substr(0, idIndex));
+            last = '#';
+            remaining = remaining.substr(idIndex + 1);
+        }
+
+    } while(idIndex !== -1 || classIndex !== -1)
+
+    parts.push(last + remaining);
+
+    return parts;
 }
--- a/virtual-hyperscript/test/h.js
+++ b/virtual-hyperscript/test/h.js
@@ -176,3 +176,22 @@ test("h with two ids", function (assert) {
 
     assert.end()
 })
+
+test("h with tag and two classes", function (assert) {
+    var node = h("div.foo.bar")
+
+    assert.equal(node.tagName, "DIV")
+    assert.equal(node.properties.className, "foo bar")
+
+    assert.end()
+})
+
+test("h with tag, id and class", function (assert) {
+    var node = h("div#foo.bar")
+
+    assert.equal(node.tagName, "DIV")
+    assert.equal(node.properties.id, "foo")
+    assert.equal(node.properties.className, "bar")
+
+    assert.end()
+})


Zolmeister:
How can we solve the problem for all use-cases? (personally I use class shorthand more)
Caching results might also help

smelnikov:
`indexOf` + `substr` will probably do the job, though the code won't be pretty.

smelnikov:
Ok, made a quick amend to my commit in order to remove browser-split all together and use a manual `indexOf` + `substr` approach. Performance seems to be unaffected from my previous commit and the class/id shorthand is now quick as well.

Raynos:
I'm :+1: for this if tests pass :)

smelnikov:
Do we want to support the case where spaces are used eg: `h('div #foo .bar .baz')`. There are no test cases for it. 

Also test cases for for the shorthand notation seem pretty slim. Only `'.foo'` `'#foo'` are tested. Wouldn't hurt to check for `'tag#foo'` and `'tag.foo'` as well.

staltz:
> Wouldn't hurt to check for 'tag#foo' and 'tag.foo' as well.

Also `'tag#foo.bar'`

neonstalwart:
@smelnikov i don't think spaces were allowed before and i don't think it should be introduced.  as for tests, `'#important.pretty'` is tested but i do agree that it wouldn't hurt to add some more tests.

Raynos:
We should test `div#tag.foo.bar` and `div.foo.bar`

smelnikov:
Actually there already is pretty decent coverage for the short hand notation but it's in `/test` and I was looking in `/virtual-hyperscript/test`. At first glance it seemed like `/test/` were for integration and `/*/test` were for unit, but the virtual-hyperscript tests in `/test` seems to be testing virtual-hyperscript in isolation.

Matt-Esch:
It's likely that some tests for the old implementation got left behind when we moved virtual-hyperscript into virtual-dom. Moving the tests into the virtual-hyperscript test directory isn't a bad idea.

smelnikov:
I've added a couple of tests to h.js that were suggested here. I ended up leaving the tests in the main test.js file alone as they depended on `Node` and I'm not familiar enough with the project to make the call as to how to restructure that. Maybe that can be done in another pull request.

Matt-Esch:
> @smelnikov i don't think spaces were allowed before and i don't think it should be introduced. as for tests, '#important.pretty' is tested but i do agree that it wouldn't hurt to add some more tests.

I'm interested in resolving this point before merging. Could you add a test for it and fix if necessary?

smelnikov:
@Matt-Esch sure. So we are not supporting spaces, correct? 

As in the previous implementation spaces were allowed and it would effectively trim them out. 

Now if you use spaces for example `h("div #foo .bar .baz ")` you will get `{tagName : 'div ', id: 'foo ', className: 'bar  baz '}` . Is this the desired behaviour that I should write a test against?

Matt-Esch:
@neonstalwart is this what you expect ^

neonstalwart:
i had thought that the current implementation did not allow spaces but it seems it does - this is unfortunate.  my expectation was that the string should be just like a CSS selector.  `.foo.bar` is a different selector to `.foo .bar` and the semantics of what we are doing matches `.foo.bar` which is why my preference is not to include spaces.  i have used a library before where `.foo .bar` would create `<div class="foo"><div class="bar"></div></div>` and in fact, the CSS selector analogy is why i think i had asked if we could include support for attributes via `.foo.bar[attr=value]` (can't find a reference to that now).

so, the question is whether or not to continue to support spaces in the selectors.

considering semver, if we continue to support them then we can increase the version number by incrementing the patch number - end of story.  if we stop supporting them then we are faced with the question of whether or not they were a part of the public API. were they possible only due to a loophole and not really a part of the public API, a claim supported by the fact that there are no tests covering this case?  the most conservative approach would be to bump the major version number if we stop supporting them but this would also be unfortunate since i'd rather not churn through major version numbers if it can be avoided.

whatever the outcome, we need to add a test with spaces to lock in the decision.

the options which i see, in order of my preference:

1. ideally i would rather not support spaces because it's inconsistent with the CSS selector parallel and i would like to be able to claim that this change should only be considered a patch increment for the version since it's closing a loophole rather than changing the public API.
1. don't support spaces and increase the major version
1. make spaces clearly part of the API and increase the patch version

Zolmeister:
There is no excuse for major version bump over this.
While possibly undocumented, it's easy to assume many inadvertently [rely on this feature](https://github.com/claydotio/clay-mobile/blob/master/src/components/dev_dashboard_contact/index.coffee#L21)
IMO there is no valid use-case to generate a nested structure via the tag name, and adding support for spaces does not hurt anything.

Spaces should clearly be part of the API

neonstalwart:
FYI @Zolmeister that link doesn't have any spaces and uses a syntax for attributes not supported by virtual-hyperscript. my main problem with allowing spaces is the inconsistency with the way CSS selectors work. by being consistent it's easy to introduce someone to this syntax. if we deviate then we increase the cognitive burden of reading, and to a lesser extent writing, our code.

Zolmeister:
@neonstalwart Agreed, but it's there already and there is no value added for clients to remove that feature. Also, spaces in the string were not a loophole. Worst case, document that using spaces may be deprecated in the next major version.

Matt-Esch:
In response to

> i'd rather not churn through major version numbers if it can be avoided.

and

> There is no excuse for major version bump over this.

Versions are cheap, you can't use semver properly and preserve branding rights for version numbers. If we make a breaking change with this we bump the major version.

Zolmeister:
@Matt-Esch My point was that we should not break backwards compatibility, not that we shouldn't bump the version.

On that note, major version changes are not cheap. We have a lot of code which depends on virtual-dom and a breaking change is expensive for us.

algesten:
@Zolmeister a breaking change is only expensive for you if you rely on that particular thing changing. A library should not be afraid of breaking changes or it will go stale.

Matt-Esch:
Ok @neonstalwart I think I am going to merge this and consider it a patch. Honestly, I consider the supported semantics to be that of CSS selectors. If spaces naturally do something then that's fine but it's undefined behavior which doesn't deserve major version consideration.

neonstalwart:
@Matt-Esch i *think* you're agreeing with me (i.e. don't support spaces in the selectors)... so, no objection.

Matt-Esch:
//TODO: I still need to land this

melnikov-s:
Hey, It's been a while ...

Any plans to merge this in? I did some benchmarking a while back on a todomvc app and this fix yielded something like a ~5% improvement overall IIRC. So not insignificant. 

Anyways, if there is a renewed interest in this PR, I can update it to fix the merge conflicts.

PR #181, opened 2015-02-06
fix insertOffset logic for node reordering

neonstalwart:
this addresses an off-by-one error introduced as part of #156

cc @AlexGalays


AlexGalays:
Thanks @neonstalwart; I won't be able to test your fix until next week though.

PR #185, opened 2015-02-09
Export dist for 1.3.0.

CRogers:
Export dist for 1.3.0. I ran `npm run dist` on top of the `v1.3.0` tag to do this.


Raynos:
lgtm.

Issue #85, opened 2014-06-21
Execution of hooks - v0.1.0


Matt-Esch:
There are outstanding questions that need a solid answer before a v0.1.0 release:
- I want to execute the hook before all properties are set
- I want to execute the hook after all properties are set
- I want to execute the hook in-order with properties (key order)
- I want to set property/execute hook immediately before adding/removing from the DOM
- I want to set property/execute hook immediately after adding/remove from the DOM
- I want to delay hook execution until I have appended the root node to the DOM
- I want to execute hooks only if the parent node is different (soft hooks)

Gozala:
I think it would be a lot easier for users if hooks had designated lifecycle methods for each of the execution state.

Issue #75, opened 2014-05-19
Hook execution guarantees


Matt-Esch:
Some work is required to ensure that hooks execute on consistent assumptions for inserts. If a new subtree is created and inserted, ideally the hooks are executed after the node has been inserted in the DOM. This means that the render process is a DOM renderer followed by an hook patch operation.

Ideally we would also be able to apply the same strategy to the root node, so it would appear that the vdom render function should return a root node and a set of patch ops for hook execution. Either this or we provide the render function with a mounting function so that we can internally apply the hooks post-mount.

Interested to hear your thoughts on this.

Gozala:
I think assumption that hooks should execute before or after node has being inserted into node is invalid. I think some hooks may run before node is in the tree while others after they are in the tree. I've being actually struggling with reacts `componentMounted / componentDidUpdate` hook a lot because in project I'm working on we depend on some non standard attributes / properties that need to be set before node is in tree or they ignored, while there are some other attributes / properties that throw if you try the set before node is in the tree.

I would very much liked if virtual-dom allowed me to do things before and after node is the tree vs one or the other.

PR #189, opened 2015-02-16
Added hook/nextTick based example of inserting element through transition

ih2502mk:


Raynos:
lgtm.

docs/css-animations.md
```diff
@@ -1,8 +1,52 @@
...
+  ]);
+}
 
-This is where the flaw in this topic is; the lack of encoding of the animation state. You don't have to do animations with JS, I don't recommed it over CSS transitions, but you do need to model your expectations properly in your data model and apply transitions to the nodes using that data. Generic transitions that rely on the way in which the DOM is mutated isn't going to work consistently.
+function ItemInsertHook(value) {
+  this.value = value;
+}
+
+ItemInsertHook.prototype.hook = function(elem, propName) {
+  if (!elem.childNodes.length) {
```
neonstalwart: am i right that this is implying that because there are no `childNodes` then this node is newly inserted?  if so, is that an assumption that may be reaching too far?  it would be relying on things happening in a certain order during `patch` which is something that is discouraged.  is there a better test like `document.contains(elem)` or something that isn't using an inference?

neonstalwart: > it would be relying on things happening in a certain order during patch

i.e it relies on hooks being applied before children are added.

ih2502mk: Yes, I agree `if (!elem.childNodes.length) {` is a quite lame way to check if the node is a newly inserted one. 
`Node.contains(elem)`, looks like a much better option.

neonstalwart: > Node.contains(elem), looks like a much better option.

you need to be careful with that too because if we are applying this hook against a node which has been re-used then the node will already be in the document.  what *really* needs to happen is that your state/view model/model (whatever you're happy to call it) needs to actually encode the fact that this node was just inserted.  it's the only reliable way to know this and that is the point of what this doc was about - you have to encode this data in your model.

ih2502mk: For the sake of this particular example, I guess `Node.contains(elem)` is fine because the code reflects the intent to check if an element is in the dom or detached. Unlike workaraoundish `!elem.childNodes.length` that does something that indirectly checks for the same condition.

But in broader perspective it's a case by case situation where you need to consider how the dom looks like what happens in reaction to what, what are the timings. And that discussion is more about hooks, their lifecycle and guarantees and it belongs to #75 and #85.

neonstalwart: fair enough.

arianvp:
How would you do transitions of removed items?

neonstalwart:
@arianvp there is some more discussion over at https://github.com/Matt-Esch/virtual-dom/issues/104#issuecomment-73732749 - but no solid answer.

PR #190, opened 2015-02-16
Added comment to code example in css transition doc 

ih2502mk:
Added comment to code example in css transition doc.
Regarding @neonstalwart comments on pr #189.


neonstalwart:
thanks

PR #194, opened 2015-02-19
fix another off-by-one logic error for reordering

### LINKED ISSUE #191: Correctness issue with keyed virtual nodes
Minimal working example: http://jsfiddle.net/CARogers/o1m76ek8/2/

In the above example there are 3 states (see `gridsOfNodesToRender`) - a 5x2 grid, followed by a 5x3 grid then a 5x2 grid. Each virtual node is keyed with a unique key for it's position in the grid (the key is also the text of each square). At the end we should be left with a 5x2 grid, but it is diffed or patched wrong so there is at least one node in the wrong place (see `2-4`):

Frame 1 - 5x2 | Frame 2 - 5x3 | Frame 3 - 5x2
--------- | ------------- | ------------
screen shot 2015-02-18 at 17 25 07 | screen shot 2015-02-18 at 17 25 17 | wrongly positioned

This is a pared down example of a real problem I encountered in production code. I made an example closer to the actual problem here: http://jsfiddle.net/CARogers/tsjLprv0/

You can drag the squares around and they'll hide if they go off the edge of the grey square. Example video: https://www.dropbox.com/s/bnxf995zpjqit33/virtualDom.mov?dl=0

neonstalwart:
fixes #191

cc @CRogers - thanks for the test case.  please try this next chance you get.  thanks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 .DS_Store
 .monitor
 .*.swp
--- a/test/keys.js
+++ b/test/keys.js
@@ -443,9 +443,9 @@ test("adding multiple widgets", function (assert) {
     assert.end()
 })
 
-test('grids of 3x2, 3x1 then 3x2', function (assert) {
-    function item(x, y) {
-        var key = x + '-' + y
+test.only('grids of 3x2, 3x1 then 3x2', function (assert) {
+    function item(key) {
+        key = key.toString()
         return h('div', {key: key}, key)
     }
 
@@ -454,31 +454,30 @@ test('grids of 3x2, 3x1 then 3x2', function (assert) {
     }
 
     var gridOf3x2 = container([
-        item(0, 0),
-        item(1, 0),
-        item(0, 1),
-        item(1, 1),
-        item(0, 2),
-        item(1, 2)
+        item(0),
+        item(1),
+        item(2),
+        item(3),
+        item(4),
+        item(5)
     ])
 
     var gridOf3x1 = container([
-        item(0, 0),
-        item(0, 1),
-        item(0, 2)
+        item(0),
+        item(2),
+        item(4)
     ])
 
     var rootNode = render(gridOf3x1)
     rootNode = patch(rootNode, diff(gridOf3x1, gridOf3x2))
-    rootNode = patch(rootNode, diff(gridOf3x2, gridOf3x1))
 
     function expectTextOfChild(childNo, text) {
         assert.equal(rootNode.childNodes[childNo].textContent, text)
     }
 
-    expectTextOfChild(0, '0-0')
-    expectTextOfChild(1, '0-1')
-    expectTextOfChild(2, '0-2')
+    for (var i = 0; i <= 5; i++) {
+        expectTextOfChild(i, i.toString())
+    }
 })
 
 function childNodesArray(node) {
--- a/test/keys.js
+++ b/test/keys.js
@@ -443,7 +443,7 @@ test("adding multiple widgets", function (assert) {
     assert.end()
 })
 
-test.only('grids of 3x2, 3x1 then 3x2', function (assert) {
+test('3 elements in a container, insert an element after each', function (assert) {
     function item(key) {
         key = key.toString()
         return h('div', {key: key}, key)
@@ -453,7 +453,13 @@ test.only('grids of 3x2, 3x1 then 3x2', function (assert) {
         return h('div', children)
     }
 
-    var gridOf3x2 = container([
+    var threeItems = container([
+        item(0),
+        item(2),
+        item(4)
+    ])
+
+    var sixItems = container([
         item(0),
         item(1),
         item(2),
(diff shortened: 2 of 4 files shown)


CRogers:
Well, the good news is the particular test case seems fixed! But it still goes wrong for me with a 5x3 grid: http://jsfiddle.net/CARogers/o1m76ek8/4/. The good news is the number of disappearing nodes in my big application has noticeably reduced, so there's probably at least one other off by one error somewhere. I'll try and make some more test cases (and try to keep them as small as possible for our sanity).

neonstalwart:
well... that's good and bad. I had a feeling there's still something lurking in there. thanks for your help.

CRogers:
I've added a further test case to https://github.com/CRogers/virtual-dom/commits/feature/191-keyed-correctness-tests (there's some code duplication for the moment but eh). Interestingly, this one is kind of the inverse of the first one - we're removing every other element rather than adding them. If we're lucky this might actually cover all the cases.

neonstalwart:
@CRogers can you try again?  i fixed your test so that it would pass but in working my way through the code paths being exercised by your test i found another optimization we can do - we were doing some reordering that wasn't necessary if we accounted for the effect of nodes being removed.  i never found any issue based on your test but i'll take an optimization as a win :smiley:.  please let me know if you're able to come up with a test case that does cause a fail - feel free to continue to adjust the one you were already using since the assertions in there for the optimizations i made are now covered by other tests so you can completely do what you want with that test case.

@Matt-Esch i've adjusted the diff code so that reordering takes into account the index of what a node would be after removals have been applied.  by doing this, i was able to completely avoid doing a reorder during patch for cases that were just removals.  please take a look.

CRogers:
Sorry about the bad test case (shouldn't code when tired). Sadly still looks borked, I added another couple of test cases at https://github.com/CRogers/virtual-dom/commits/feature/191-keyed-correctness-tests. Unfortunately they're a bit bigger but the only ones I could find that failed.

Matt-Esch:
@neonstalwart I am going to pay close attention to this today. This is fundamental and can't be broken.

neonstalwart:
I agree. my plan is to see if we can add some fuzzy testing that provides many more test cases than we could write by hand.

CRogers:
Yeah, I basically found those two test cases by manually trying many combinations of the constants. This is something that could be automated really easily.

neonstalwart:
FYI - I'm not currently working on this and don't expect I'll have time until next week. so if either of you have time this weekend then go ahead.

Matt-Esch:
So I verified @CRogers tests on his feature/191-keyed-correctness-tests branch

Most of the problems with the tests are just bad tests. However after refactoring, I think there is a genuine issue with the last test.

It generates 10 items with keys `[0, 1, 3, 4, 6, 7, 9, 10, 12, 13]`. (Every third is missing). We then attempt to patch a full set of `[0,..,14]` onto it. The patch contains 5 inserts for `[2, 5, 8, 11, 14]` followed by a final order patch.

The contents of the order patch are

```json
{
    "2": 10,
    "3": 2,
    "4": 3,
    "5": 11,
    "6": 4,
    "7": 5,
    "8": 12,
    "9": 6,
    "10": 7,
    "11": 13,
    "12": 8,
    "13": 9,
    "removes": {},
    "reverse": {
        "2": 3,
        "3": 4,
        "4": 6,
        "5": 7,
        "6": 9,
        "7": 10,
        "8": 12,
        "9": 13,
        "10": 2,
        "11": 5,
        "12": 8,
        "13": 11
    }
}
```

And the issue is that item 14 is out of place. So to run through the indices again

We start with

`[0, 1, 3, 4, 6, 7, 9, 10, 12, 13]`

We insert the 5 missing nodes at the end

`[0, 1, 3, 4, 6, 7, 9, 10, 12, 13, 2, 5, 8, 11, 14]`

We apply the final reorder and end up with

`[0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 14, 12, 13]`

neonstalwart:
ok, so if everything worked, we should see the `insertBefore` call happen with the following pairs of nodes

1. 2, 3
1. 5, 6
1. 8, 9
1. 11, 12
1. 14, null (or maybe not even at all)

this smells like some more bad accounting with `insertOffset`

Matt-Esch:
A trace of the moves that actually happen (read this as insert <id> before <id>)

```
insert 2 before 3
insert 3 before 4
insert 5 before 7
insert 6 before 9
insert 7 before 9
insert 8 before 12
insert 9 before 13
insert 10 before 13
insert 11 before 14
insert 12 before null
insert 13 before null
```

CRogers:
Mmm, those `null`s would explain all the "can't call function insertBefore on undefined" errors I keep getting. Did either of you come up with any explanation about what might be happening?

neonstalwart:
insertBefore null is not a problem - it means, add to the end of the list of childNodes - http://www.w3.org/TR/2000/REC-DOM-Level-2-Core-20001113/core.html#ID-952280727

reordering is being rewritten.  the logic is moving to the diff phase so that the patch phase gets a list of moves like:

```js
[
  { from: 6, to: 5 },
  { from: 4, to: 8 }
]
```

neonstalwart:
with this rewrite there will be more tests... e.g. https://github.com/neonstalwart/virtual-dom/commit/f980c558bb43a10a9889d9ab07cd60f754375d05

we (especially me) don't want to get this wrong again - please have patience while we use what little time we have to make this right.

CRogers:
Oh excellent (I didn't notice the new branch) - thanks for the all the effort you're putting into this :)

Issue #198, opened 2015-02-25
function reorderChildren in patch-op.js fails in certain cases 


yiransheng:
See this jsbin:
[http://jsbin.com/hucuca/1/edit?html,js,console,output](http://jsbin.com/hucuca/1/edit?html,js,console,output)

`reorderChildren` function from: [here](https://github.com/Matt-Esch/virtual-dom/blob/86dc068cc4cd04a9b984c9a21cd7aaca53e038e4/vdom/patch-op.js) throws errors for certain edge cases. For example, in the jsbin link above, when calling `reorderChildren` on a DOM node with 43 childNodes (all DIVs), with new ordering: `[41, 3, 34, 36, 1, 40, 39, 7, 37, 14, 23, 26, 15, 6, 25, 24, 19, 8, 9, 22, 29, 27, 38, 35, 11, 20, 33, 31, 17, 32, 4, 28, 12, 2, 10, 0, 42, 21, 5, 16, 30, 18, 13]` (easily verified as a permutation of [0, 1, 2, ... 42]),  this error gets thrown:

```
TypeError: Cannot use 'in' operator to search for '0' in undefined
```

The result DOM node is not fully reordered according to input/patch, this error happens rarely, but will accumulate overtime, and cause real DOM and virtual-dom to be out of sync. 

EDIT: Ubuntu 14.04 Chrome Version 40.0.2214.115 (64-bit)

Matt-Esch:
This should be fixed in v2.0.0 - reopen the issue if not

PR #199, opened 2015-02-27
Rewrite reorder (part 2)

### LINKED ISSUE #191: Correctness issue with keyed virtual nodes

neonstalwart:
@Matt-Esch this is some extra work on top of what you've done in #197.  please take the time to convince yourself that this is reasonable - i don't want to get this wrong again. i have relied on tests to prove this code as being correct but that's only as good as the tests themselves.

things of note:

* added some fuzzy testing - *please* check what i've done with the tests and make sure it's reasonable because i'm depending on these tests to make sure things are working.  there are some specific cases also included to exercise some corner cases and to make debugging/development easier.
* allowed for left-to-right moves when reordering.  this means that a change that moves one key from position 4 to position 7 is now just `[{ from: 4, to: 7}]` rather than `[{ from: 5, to: 4 }, { from: 6, to: 5 }, { from: 7, to: 6 }]`
* removed `while(!sorted)` loop which i think was previously making a reorder O(N<sup>3</sup>) (O(N) for the while loop * O(N) for the for loop * O(N) for searching for the node to move) and now it's O(N<sup>2</sup>) (O(N) for the for loop * O(N) for searching for the node to move)

this PR supersedes or fixes #191, #194, #197, #198

as i mentioned in #197, i think this change warrants a major version bump since the output format of a diff is changing in a backwards incompatible way (but i think it's definitely the right thing to do).  in doing that, let's look at what else may be something to address in a major version bump (e.g. stop allowing whitespace in selectors as discussed in #180)

### DIFF
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -134,11 +134,12 @@ function reorderChildren(domNode, moves) {
         domNode.removeChild(node)
     }
 
+    var length = childNodes.length
     for (var j = 0; j < moves.inserts.length; j++) {
         insert = moves.inserts[j]
         node = keyMap[insert.key]
         // this is the weirdest bug i've ever seen in webkit
-        domNode.insertBefore(node, insert.to >= childNodes.length ? null : childNodes[insert.to])
+        domNode.insertBefore(node, insert.to >= length++ ? null : childNodes[insert.to])
     }
 }
 
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -120,20 +120,25 @@ function destroyWidget(domNode, w) {
 
 function reorderChildren(domNode, moves) {
     var childNodes = domNode.childNodes
+    var keyMap = {}
+    var node
+    var remove
+    var insert
 
-    for (var i = 0; i < moves.length; i++) {
-        var move = moves[i]
-        var from = move.from;
-        var to = move.to;
-
-        if (to >= 0) {
-            domNode.insertBefore(
-                childNodes[move.from],
-                childNodes[move.to]
-            )
-        } else {
-            domNode.removeChild(childNodes[from])
+    for (var i = 0; i < moves.removes.length; i++) {
+        remove = moves.removes[i]
+        node = childNodes[remove.from]
+        if (remove.key) {
+            keyMap[remove.key] = node
         }
+        domNode.removeChild(node)
+    }
+
+    for (var j = 0; j < moves.inserts.length; j++) {
+        insert = moves.inserts[j]
+        node = keyMap[insert.key]
+        // this is the weirdest bug i've ever seen in webkit
+        domNode.insertBefore(node, insert.to >= childNodes.length ? null : childNodes[insert.to])
     }
 }
 
--- a/vtree/diff.js
+++ b/vtree/diff.js
@@ -332,23 +332,20 @@ function reorder(aChildren, bChildren) {
                         // if the remove didn't put the wanted item in place, we need to insert it
                         if (!simulateItem || simulateItem.key !== wantedItem.key) {
                             inserts.push({key: wantedItem.key, to: k})
-                            k++
                         }
                         // items are matching, so skip ahead
                         else {
                             simulateIndex++
-                            k++
                         }
                     }
                     else {
                         inserts.push({key: wantedItem.key, to: k})
-                        k++
                     }
                 }
                 else {
                     inserts.push({key: wantedItem.key, to: k})
-                    k++
                 }
+                k++
             }
             // a key in simulate has no matching wanted key, remove it
             else if (simulateItem && simulateItem.key) {
--- a/vtree/diff.js
+++ b/vtree/diff.js
@@ -110,7 +110,7 @@ function diffChildren(a, b, patch, apply, index) {
         }
     }
 
-    if (orderedSet.moves.length > 0) {
+    if (orderedSet.moves) {
         // Reorder nodes last
         apply = appendPatch(apply, new VPatch(
             VPatch.ORDER,
@@ -224,12 +224,11 @@ function reorder(aChildren, bChildren) {
     var bChildIndex = keyIndex(bChildren)
     var bKeys = bChildIndex.keys
     var bFree = bChildIndex.free
-    var moves = []
 
     if (bFree.length === bChildren.length) {
         return {
             children: bChildren,
-            moves: moves
+            moves: null
         }
     }
 
@@ -241,13 +240,12 @@ function reorder(aChildren, bChildren) {
     if (aFree.length === aChildren.length) {
         return {
             children: bChildren,
-            moves: moves
+            moves: null
         }
     }
 
     // O(MAX(N, M)) memory
     var newChildren = []
-    var sortOrder = []
 
     var freeIndex = 0
     var freeCount = bFree.length
@@ -264,36 +262,23 @@ function reorder(aChildren, bChildren) {
                 // Match up the old keys
                 itemIndex = bKeys[aItem.key]
                 newChildren.push(bChildren[itemIndex])
-                sortOrder.push(itemIndex)
 
             } else {
                 // Remove old keyed items
-                // Moving an item to -1 deletes it. We have to keep
-                // track of items have been deleted as all items are shifted
-                // to the left after removal
                 itemIndex = i - deletedItems++
                 newChildren.push(null)
-                moves.push({
-                    from: itemIndex,
-                    to: -1
-                })
             }
         } else {
             // Match the item in a with the next free item in b
             if (freeIndex < freeCount) {
                 itemIndex = bFree[freeIndex++]
                 newChildren.push(bChildren[itemIndex])
-                sortOrder.push(itemIndex)
             } else {
                 // There are no free items in b to match with
                 // the free items in a, so the extra free nodes
                 // are deleted.
                 itemIndex = i - deletedItems++
                 newChildren.push(null)
-                moves.push({
-                    from: itemIndex,
-                    to: -1
-                })
             }
         }
     }
@@ -313,82 +298,100 @@ function reorder(aChildren, bChildren) {
                 // We are adding new items to the end and then sorting them
                 // in place. In future we should insert new items in place.
                 newChildren.push(newItem)
-                sortOrder.push(j);
             }
         } else if (j >= lastFreeIndex) {
             // Add any leftover non-keyed items
             newChildren.push(newItem)
-            sortOrder.push(j)
         }
     }
 
-    sortIndex(sortOrder, moves) // O(N^2) worst case
+    var simulate = newChildren.slice()
+    var simulateIndex = 0
+    var removes = []
+    var inserts = []
+    var simulateItem
+
+    for (var k = 0; k < bChildren.length;) {
+        var wantedItem = bChildren[k]
+        simulateItem = simulate[simulateIndex]
+
+        // remove items
+        while (simulateItem === null && simulate.length) {
+            removes.push(remove(simulate, simulateIndex, null))
+            simulateItem = simulate[simulateIndex]
+        }
+
+        if (!simulateItem || simulateItem.key !== wantedItem.key) {
+            // if we need a key in this position...
+            if (wantedItem.key) {
+                if (simulateItem && simulateItem.key) {
+                    // if an insert doesn't put this key in place, it needs to move
+                    if (bKeys[simulateItem.key] !== k + 1) {
+                        removes.push(remove(simulate, simulateIndex, simulateItem.key))
+                        simulateItem = simulate[simulateIndex]
+                        // if the remove didn't put the wanted item in place, we need to insert it
+                        if (!simulateItem || simulateItem.key !== wantedItem.key) {
+                            inserts.push({key: wantedItem.key, to: k})
+                            k++
+                        }
+                        // items are matching, so skip ahead
+                        else {
+                            simulateIndex++
+                            k++
+                        }
+                    }
+                    else {
+                        inserts.push({key: wantedItem.key, to: k})
+                        k++
+                    }
+                }
+                else {
+                    inserts.push({key: wantedItem.key, to: k})
+                    k++
+                }
+            }
+            // a key in simulate has no matching wanted key, remove it
+            else if (simulateItem && simulateItem.key) {
+                removes.push(remove(simulate, simulateIndex, simulateItem.key))
+            }
+        }
+        else {
+            simulateIndex++
+            k++
+        }
+    }
+
+    // remove all the remaining nodes from simulate
+    while(simulateIndex < simulate.length) {
+        simulateItem = simulate[simulateIndex]
+        removes.push(remove(simulate, simulateIndex, simulateItem && simulateItem.key))
+    }
 
     // If the only moves we have are deletes then we can just
     // let the delete patch remove these items.
-    if (moves.length === deletedItems) {
+    if (removes.length === deletedItems && !inserts.length) {
         return {
             children: newChildren,
-            moves: []
+            moves: null
         }
     }
 
     return {
         children: newChildren,
-        moves: moves
-    }
-}
-
-function sortIndex(arr, moves) {
-    var arrLength = arr.length;
-
-    for (var i = 0; i < arrLength; i++) {
-        var item = arr[i];
-
-        if (item !== i) {
-            var desiredItemIndex = 0;
-            var swapDestination = 0;
-            var lessThanSearchHalt = item - i;
-            var lessThanSearchCount = 0;
-
-            for (var j = i + 1; j < arrLength; j++) {
-                var compareItem = arr[j];
-
-                if (compareItem < item) {
-                    lessThanSearchCount += 1;
-                    if (lessThanSearchCount === lessThanSearchHalt) {
-                        swapDestination = j + 1;
-                    }
-                }
-
-                if (compareItem === i) {
-                    desiredItemIndex = j;
-                }
-
-                if (desiredItemIndex > 0 && swapDestination > 0) {
-                    break;
-                }
-            }
-
-            if (swapDestination > desiredItemIndex + 1) {
-                moves.push(move(arr, i, swapDestination));
-                i--;
-            } else {
-                moves.push(move(arr, desiredItemIndex, i));
-            }
+        moves: {
+            removes: removes,
+            inserts: inserts
         }
     }
-
-    return moves
 }
 
-function move(array, from, to) {
-    array.splice(from < to ? to - 1 : to, 0, array.splice(from, 1)[0]);
+function remove(arr, index, key) {
+    arr.splice(index, 1)
 
     return {
-        from: from,
-        to: to
-    };
+        from: index,
+        key: key
+    }
 }
 
 function keyIndex(children) {
--- a/test/keys.js
+++ b/test/keys.js
@@ -13,27 +13,8 @@ var assertChildNodesFromArray = require("./lib/assert-childNodes-from-array.js")
 var VPatch = require("../vnode/vpatch.js")
 
 test("keys get reordered", function (assert) {
-    var leftNode = h("div", [
-        h("div", { key: 1 }, "1"),
-        h("div", { key: 2 }, "2"),
-        h("div", { key: 3 }, "3"),
-        h("div", { key: 4 }, "4"),
-        h("div", { key: "test" }, "test"),
-        h("div", { key: 6 }, "6"),
-        h("div", { key: "good" }, "good"),
-        h("div", { key: "7" }, "7")
-    ])
-
-    var rightNode = h("div", [
-        h("div", { key: "7" }, "7"),
-        h("div", { key: 4 }, "4"),
-        h("div", { key: 3 }, "3"),
-        h("div", { key: 2 }, "2"),
-        h("div", { key: 6 }, "6"),
-        h("div", { key: "test" }, "test"),
-        h("div", { key: "good" }, "good"),
-        h("div", { key: 1 }, "1")
-    ])
+    var leftNode = nodesFromArray(["1", "2", "3", "4", "test", "6", "good", "7"])
+    var rightNode = nodesFromArray(["7", "4", "3", "2", "6", "test", "good", "1"])
 
     var rootNode = render(leftNode)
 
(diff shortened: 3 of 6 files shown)


vtree/diff.js
```diff
@@ -343,16 +348,23 @@ function reorder(aChildren, bChildren) {
 // TODO: We really want to make this more efficient
 // Finding the current index of any key in the array could be
 // faster.
-function moveItem(array, toIndex, itemKey) {
+function indexOfKey(array, itemKey) {
```
Zolmeister: Can't we use a hash for this? Reducing the complexity to O(N) would be great.

Don't do this, but a hack idea:
```js
function indexOfKey(array, itemKey) {
  if (array.keyMap) return array.keyMap[itemKey]
  for (i in array) {
    array.keyMap[array[i].key] = i
  }
return array.keyMap[itemKey]
```
And it's updated on array mutation

neonstalwart: the updating after a mutation is O(N) which is the same as not keeping the hash and scanning the list like we do now - the hash is technically worse (by just a little bit) since it requires more memory.

neonstalwart: i.e. scanning the list is already O(N)

Zolmeister: Updated: just run once before a patch cycle

Zolmeister: Well, that, and keeping it updated incrementally

neonstalwart: the list mutates during the patch so it has to update after each mutation... which is no more efficient than looking for the item each time we mutate (like we do now).

...and to cut off another thought, `array.indexOf` is likely to perform just the same as what we're doing but only some performance tests can really prove if it's faster.

neonstalwart: > keeping it updated incrementally

the incremental updates are O(N)... still the same.  i've given this a lot of thought this week :smile:

vtree/diff.js
```diff
@@ -288,50 +288,55 @@ function reorder(aChildren, bChildren) {
...
-                simulateItem = simulate[++simulateIndex]
-            }
+        // we don't have a match
+        if (!simulateItem || simulateItem.key !== wantedItem.key) {
+            if (wantedItem.key) {
+                var move = null
+                // item at current location moves left to right later
+                if (simulateItem && simulateItem.key && bKeys[simulateItem.key] > k) {
+                    // only do the current move if it is left to right
+                    if (indexOfKey(simulate, wantedItem.key) < k) {
```
Zolmeister: index could be cached here, and passed into moveItem to avoid another lookup

vtree/diff.js
Zolmeister: I'm sure this is possible. React does it in O(N) with keys: http://facebook.github.io/react/docs/reconciliation.html
I'll think about it more.

neonstalwart: i just took a look at what react does and we could achieve O(N) if we were to sacrifice the same things they have given up.

take this example:

```jsx
React.render(
    <ul>
      <li key="1" id="1">1</li>
      <li key="2" id="2">2</li>
      <li key="3" id="3">3</li>
      <li key="0" id="0">0</li>
      <li key="4" id="4">4</li>
  </ul>,
  document.getElementById('example')
);
                                         
setTimeout(function () {
  React.render(
      <ul>
          <li key="0" id="0">0</li>
          <li key="1" id="1">1</li>
          <li key="2" id="2">2</li>
          <li key="3" id="3">3</li>
          <li key="4" id="4">4</li>
      </ul>,
      document.getElementById('example')
  );
}, 500)
```

intuitively there should be one move to get the element with `key="0"` into it's final position (from 3, to 0).  however, react instead produces 3 moves (from 0, to 1; from 1, to 2; from 2, to 3).  moving all these nodes around has some cost associated with it and will trigger blur handlers and break animations on those nodes.  the question is then whether the cost of calculating the efficient moves outweighs the cost (and breaking developer's expectations of nodes remaining in the DOM) of moving lots of nodes in the DOM.  intuitively i believe that for a typical number of keyed nodes (a low number), calculating the efficient patch and keeping more nodes in the DOM more often is the better option but we don't have any conclusive performance metrics to back this up.  http://vdom-benchmark.github.io/vdom-benchmark/ seems to suggest that so far virtual-dom has produced generally significantly better performance than react.

i'd really like to be able to get O(N) working *AND* produce the same set of moves this code currently produces - i know it's possible but for now i'm favoring working code over the most performant code.  when we can get the performant code to be robust then i think we should switch to it.

vtree/diff.js
neonstalwart: done. thanks.

Matt-Esch:
@Zolmeister of course we could make this O(N) if we were to relax the conditions for the moves. We used to have an O(N) algorithm but it did not minimize swaps at all.

We could create a mapping of desired index to dom node in patch an O(N) operation requiring O(N) ish memory depending on how the hash is defined, and then scan left to right. If a node is not in place, use the hash to insertBefore that node.

That's a single left to right iteration with N comparisons, forcing each item into the correct place. As we already noted though, there is an unfortunate amount of pain to suffer when an item at the start needs to be at the end, because it gets shuffled with N-1 swaps to the end, when we could reduce it to just N swaps. The O(N^2) algorithm with 1 swap for small enough N is likely (but hasn't been proven to be) faster overall than the O(N) algorithm with N-1 swaps. We are assuming that rearranging the DOM nodes is an expensive operation, and the JS reordering algorithm operates over cheap to mutate data structures.

Zolmeister:
@Matt-Esch Thanks for explaining, definitely makes more sense now.

Matt-Esch:
@neonstalwart I am going to take a good amount of time to read your diff and tests in particular. In the mean time I was wondering how you felt about this idea:

The previous steps before the simulation allow us to create an array of integers from 1 .. N. The array will be in the out-of-order version. So in the case that the first item gets moved to the end you might have an array of

```
[5, 0, 1, 2, 3, 4]
```

Note that sorting an array of integers is going to be more efficient than sorting an array of complex objects, and for sufficiently small N we can efficiently move items by shuffling the items manually instead of using splice...

Anyway, the algorithm I was thinking of would look at each item left to right. When it sees an item out of place it determines 2 things, a) what is the current index of the item that should be here. b) what is the index that this current item should be at.

a) is computed simply by finding the item
b) b is computed by counting enough items less than the current item.

We can only do be because we have a strict sequence from 0 ... N. That is if Item at index 3 is 7, I know that I must place 7 after 3, 4, 5, and 6.

We then either move the item that should be here to this index, OR we move the item that is currently here to the index it should be at. We make the largest move.

Here are the moves from my example algorithm:
```
sorting        [5,0,1,2,3,4]
move 0 to 5 => [0,1,2,3,4,5]
done


sorting        [1,2,3,4,5,0]
move 5 to 0 => [0,1,2,3,4,5]
done


sorting        [5,4,3,2,1,0]
move 5 to 0 => [0,5,4,3,2,1]
move 5 to 1 => [0,1,5,4,3,2]
move 5 to 2 => [0,1,2,5,4,3]
move 5 to 3 => [0,1,2,3,5,4]
move 5 to 4 => [0,1,2,3,4,5]
done


sorting        [0,1,2,3,4,5]
done


sorting        [3,4,5,0,1,2]
move 0 to 5 => [4,5,0,1,2,3]
move 0 to 5 => [5,0,1,2,3,4]
move 0 to 5 => [0,1,2,3,4,5]
done
```

And the annotated source + log statements:

```js
// arr must be an (unordered) set of integers 0, 1, 2, ... N without gaps
function sort(arr) {
    var arrLength = arr.length;
    var moves = [];
    console.log('sorting       ', JSON.stringify(arr));

    for (var i = 0; i < arrLength; i++) {
        var item = arr[i];

        if (item !== i) {
            // This is the index of the item that should be at i
            var desiredItemIndex = 0;

            // This is the index that the current item should be at
            var swapDestination = 0;

            // So if i is currently 3, say, and the item at 3 is 7, we want to
            // see 4, 5 and 6 first. So we want to see (item - i - 1) items
            // which are less than 7 to the right of the current item.
            var lessThanSearchHalt = item - i - 1;
            var lessThanSearchCount = 0;
            

            for (var j = i + 1; j < arrLength; j++) {
                var compareItem = arr[j];

                if (compareItem < item) {
                    lessThanSearchCount += 1;
                    
                    if (lessThanSearchCount === lessThanSearchHalt) {
                        swapDestination = j + 1;
                    }
                }

                if (compareItem === i) {
                    desiredItemIndex = j;
                }

                if (desiredItemIndex > 0 && swapDestination > 0) {
                    break;
                }
            }

            // Make the largest move
            if (swapDestination > desiredItemIndex) {
                moves.push(move(arr, i, swapDestination));
                // We still need to put the correct item at i.
                i--;
            } else {
                moves.push(move(arr, desiredItemIndex, i));
            }
        }
    }

    console.log('done\n\n')
    return moves;
}

// TODO: shuffle items for small enough N based on benchmarks.
function move(arr, from, to) {
    arr.splice(to, 0, arr.splice(from, 1)[0]);
    console.log("move", from, "to", to, '=>', JSON.stringify(arr));
    return {
        from: from,
        to: to
    };
}

sort([5, 0, 1, 2, 3, 4]);
sort([1, 2, 3, 4, 5, 0]);
sort([5, 4, 3, 2, 1, 0]);
sort([0, 1, 2, 3, 4, 5]);
sort([3, 4, 5, 0, 1, 2]);
```

Matt-Esch:
Also another thing to bear in mind is that for short arrays we can actually cache the moves in a tree based on the numeric ordering!

Zolmeister:
Based on [this solution](http://stackoverflow.com/questions/10763344/minimum-number-of-operations-to-make-an-array-sorted) here is a method to calculate the optimal number of moves required. Hopefully this can make it into the tests:

```js
function minMoves(arr) {
  var longestSeq = 0

  for (var i = 0; i < arr.length; i++) {
    var cnt = 1
    var val = arr[i]
    for (var j = i + 1; j < arr.length; j++) {
      if (val + 1 === arr[j]) {
        val += 1
        cnt += 1
      }
    }
    longestSeq = Math.max(longestSeq, cnt)
  }

  return arr.length - longestSeq
}

// [1, 7, 6, 2, 5, 4, 3] - 4
// [1, 6, 4, 3, 5, 2, 7] - 5
```

neonstalwart:
@Matt-Esch i need to rebase this against your updated rewrite-reorder branch (probably tomorrow, i'm mostly out of time today) but take a look at what i've done with reordering now... it's O(N)!!!

as you've mentioned elsewhere, it overlaps with insertion and deletion patches but we can address that separately

neonstalwart:
@Matt-Esch this has been rebased (and worked around a weird `childNodes[childNodes.length]` bug in webkit).  i tried making the `simulate` array just an array of keys instead of objects to see if the performance improved but i didn't see anything significant.

on the topic of performance, here's what i've been doing:
* clone https://github.com/vdom-benchmark/vdom-benchmark-virtual-dom
* `npm install` its dependencies and `npm link virtual-dom` into it so you can test your local copy
* run `gulp serve` in the root of that repo
* visit http://vdom-benchmark.github.io/vdom-benchmark/ and put in a custom URL that points to the server started in the previous step

NOTE: you only get to see the most recent output for any test with the same label.  change https://github.com/vdom-benchmark/vdom-benchmark-virtual-dom/blob/19b55f15d409af152d03bb2aecee58235e423fc6/web/js/main.js#L11 any time you want to do a side-by-side comparison with a different test run.  e.g. i had values like `'origin/master'`, `'origin/rewrite-reorder'`, `'neonstalwart/rewrite-reorder'`, `'local'`, etc. so that i could run the benchmark with different code and compare them side-by-side.  as an aside, i wish you could manipulate the name with the queryString so that you could simply put in a URL like http://localhost:3000/?name=origin/rewrite-reorder

Matt-Esch:
Didn't get a chance to look at this, review coming up.

vdom/patch-op.js
```diff
@@ -120,20 +120,25 @@ function destroyWidget(domNode, w) {
...
+        if (remove.key) {
+            keyMap[remove.key] = node
         }
+        domNode.removeChild(node)
+    }
+
+    for (var j = 0; j < moves.inserts.length; j++) {
+        insert = moves.inserts[j]
+        node = keyMap[insert.key]
+        // this is the weirdest bug i've ever seen in webkit
```
Matt-Esch: it's probably undefined behavior. Maybe there is a better way than checking the child node length? Don't really want to be checking a host object property in a tight loop when we knew upfront that it was an append operation.

Matt-Esch:
I think this algorithm suffers a subtle but important side effect in that it removes visible items (keyed items that are out of place and don't satisfy the lookahead by 1 condition). Surely this is going to cause the weird rendering hiccups we want to avoid?

I don't trust the benchmarks for testing the visible rendering time and I don't trust this as being free from undesirable rendering artifacts.

vdom/patch-op.js
neonstalwart: i'll improve this to avoid referencing `childNodes.length` - i get the reference once before the loop and then maintain my own counter

neonstalwart:
> One thing I've noticed is that this algorithm removes items that are visible. At first I thought this was terrible but I'm not sure it really is.

do you have any examples of where something is happening that should be avoided?  it should be simple enough to add a test to ensure we have what is desirable.

there will always be some ambiguous cases where an existing node has to move and we need to arbitrarily choose one node to move instead of another.

for example, if we start with `[0,1,2,4,3,5]` and we see a change to `[0,1,2,3,4,5]`, a user could have intended either of:
* move 4 after 3 - the expectation might be that 4 would be the element leaving the DOM
* move 3 before 4 - the expectation might be that 3 would be the element leaving the DOM

we don't have enough information to know which one is intended and so we leave it to the developer of the code to manage transitions etc to match the intent of the user and we just make sure that the final state of the DOM is consistent with what we know.

neonstalwart:
> Essentially the only difference between the algorithm I proposed and this one is that you're using a remove operation as a stack instead of eagerly inserting the node into the correct position.

by using the stack i can do the shuffle in O(N) (worst case is N+M, i.e. bChildren.length + simulate.length but that is still called O(N)) compared to O(N<sup>2</sup>) worst case, or O(NlogN) average case, for yours.

the other advantage is that my patch format lends itself to folding all removes and inserts into VPatch.ORDER to remove the redundancy with VPatch.INSERT and VPatch.REMOVE but this is not as big a deal as being able to reduce the big-O complexity.

neonstalwart:
> FWIW looking at an array of numbers is way easier to debug,

i worked on a variation where `simulate` was an array of keys rather than an array of vnodes.  it didn't seem to change the performance so i didn't do it but if you think it helps debugging, it's fairly trivial to switch to that.

Matt-Esch:
It's also worth noting that objects are not really hashes in javascript and that v8 does a linear lookup on the object keys to find out where in memory it's located. I don't consider using a hash as O(1) in most cases and in the v8 case it's actually an O(N) lookup.

Zolmeister:
@Matt-Esch Can you provide a link to that info?

Matt-Esch:
http://thlorenz.github.io/talks/demystifying-v8/talk.pdf

So the linear lookup is worst case. v8 "hashes" are objects in dictionary mode. This means that when you write a property, the string is hashed and first 2 bytes used to compute the index. From then on, we linearly search for the result looping back round like any other hash table implementation. This has worst case complexity of O(N).

They call objects in dictionary mode "slow mode" for a reason. This is probably too low level and v8 specific, but I do get frustrated when people think hash tables are O(1). It's only an O(1) lookup when you can go from string to offset without searching.

Zolmeister:
My bad, I forgot it requires a perfect-hash function to get O(1). (even the [wikipedia](http://en.wikipedia.org/wiki/Hash_table) page says O(n))
That being said, average case is still O(1), so we can both be right. (well, partially)

neonstalwart:
i think we all want the "best" implementation and i don't mind if that's not mine but it looks like we need to do some profiling.  ideally we want tests that include patching the DOM but i'm stretched for time right now so maybe i can start by adjusting test/sort.js to do a diff using vtree/diff rather than just copying/pasting sort algorithms in that file.  then we can at least profile both approaches in a way that is meaningful for comparison - although without patching the DOM it is not completely conclusive.  after that maybe it won't take a lot of effort to add patching the DOM to that.

apart from performance, i think there were some characteristics of your approach which were undesirable.  e.g. reordering nodes that will be removed but given the number of changes we've gone through in the past couple of weeks, i can't remember whether this is still the case or not so forgive me if i'm wrong.

neonstalwart:
using the changes to test/sort.js added in d121124 from this branch and the below config.  i got the log output from chrome included in the table below.  (all the permutations were sorted correctly but i've removed the logging output related to that).

@Matt-Esch these numbers below show no significant performance difference between the 2 approaches. it would be good to double-check that the changes to test/sort.js are reasonable and verify my results by running the tests again but i'm happy to bring this to a conclusion any way you want to - i.e. i don't care too much which code you merge as long as we aren't compromising on our other constraints such as disturbing the DOM as little as possible.

```js
var PERMUTATION_START = 1;
var PERMUTATION_END = 8;

var SAMPLE_START = 10;
var SAMPLE_END = 1000;
var SAMPLE_COUNT = 100;
var SAMPLE_INTERVAL = 100;
```


| d121124 | 9c09d58 |
|---------|---------|
| All (1) arrays sorted in 1 ms               | All (1) arrays sorted in 0 ms                 |
| An array of length 1 sorts in 1 ms          | An array of length 1 sorts in 0 ms            |
| All (2) arrays sorted in 3 ms               | All (2) arrays sorted in 2 ms                 |
| An array of length 2 sorts in 1 ms          | An array of length 2 sorts in 1 ms            |
| All (6) arrays sorted in 7 ms               | All (6) arrays sorted in 9 ms                 |
| An array of length 3 sorts in 1 ms          | An array of length 3 sorts in 1 ms            |
| All (24) arrays sorted in 38 ms             | All (24) arrays sorted in 34 ms               |
| An array of length 4 sorts in 1 ms          | An array of length 4 sorts in 1 ms            |
| All (120) arrays sorted in 115 ms           | All (120) arrays sorted in 154 ms             |
| An array of length 5 sorts in 0 ms          | An array of length 5 sorts in 1 ms            |
| All (720) arrays sorted in 749 ms           | All (720) arrays sorted in 768 ms             |
| An array of length 6 sorts in 1 ms          | An array of length 6 sorts in 1 ms            |
| All (5040) arrays sorted in 5824 ms         | All (5040) arrays sorted in 6111 ms           |
| An array of length 7 sorts in 1 ms          | An array of length 7 sorts in 1 ms            |
| All (40320) arrays sorted in 55557 ms       | All (40320) arrays sorted in 52125 ms         |
| An array of length 8 sorts in 1 ms          | An array of length 8 sorts in 1 ms            |
| All (100) arrays sorted in 143 ms           | All (100) arrays sorted in 125 ms             |
| An array of length 10 sorts in 1 ms         | An array of length 10 sorts in 1 ms           |
| All (100) arrays sorted in 1120 ms          | All (100) arrays sorted in 1201 ms            |
| An array of length 110 sorts in 11 ms       | An array of length 110 sorts in 12 ms         |
| All (100) arrays sorted in 2204 ms          | All (100) arrays sorted in 2166 ms            |
| An array of length 210 sorts in 22 ms       | An array of length 210 sorts in 21 ms         |
| All (100) arrays sorted in 3055 ms          | All (100) arrays sorted in 3078 ms            |
| An array of length 310 sorts in 30 ms       | An array of length 310 sorts in 30 ms         |
| All (100) arrays sorted in 4079 ms          | All (100) arrays sorted in 4110 ms            |
| An array of length 410 sorts in 40 ms       | An array of length 410 sorts in 41 ms         |
| All (100) arrays sorted in 5013 ms          | All (100) arrays sorted in 5058 ms            |
| An array of length 510 sorts in 50 ms       | An array of length 510 sorts in 50 ms         |
| All (100) arrays sorted in 6011 ms          | All (100) arrays sorted in 6066 ms            |
| An array of length 610 sorts in 60 ms       | An array of length 610 sorts in 60 ms         |
| All (100) arrays sorted in 6981 ms          | All (100) arrays sorted in 7034 ms            |
| An array of length 710 sorts in 69 ms       | An array of length 710 sorts in 70 ms         |
| All (100) arrays sorted in 7906 ms          | All (100) arrays sorted in 8073 ms            |
| An array of length 810 sorts in 79 ms       | An array of length 810 sorts in 80 ms         |
| All (100) arrays sorted in 9069 ms          | All (100) arrays sorted in 9537 ms            |
| An array of length 910 sorts in 90 ms       | An array of length 910 sorts in 95 ms         |

Zolmeister:
Can you output the number of moves each?

neonstalwart:
> Can you output the number of moves each?

we have other tests for ensuring the number of moves are minimal. admittedly, those tests are not comprehensive but i'm open to a PR that improves on what we have.  i've probably spent more than 60 hours on this over the past 2 weeks - my resources have a limit.

Matt-Esch:
@neonstalwart all of the hard work appreciated as always. I have a test in the works that I am going to run through some profiling which will give us some indication of any performance overheads.

My main concern is the number of DOM operations performed. I was assuming that the algorithms have a similar performance profile but that the O(N) algorithm you suggested increases the number of DOM operations (an increase in removes though I don't expect more moves).

I am concerned that this will have a performance penalty in some browser we care about which outweighs any advantage of the O(N) algorithm. I want to stop using conjecture and actually profile this in the browser and land something. Afterall, sorting is completely broken on master right now, which itself outweighs the need to squeeze every last drop.

So, I will post chrome profiles and I will time how long the JS blocks for. I'll create a test and post the results for an array of browsers, and we will just pick the one that provides the best overall performance i.e. the one that gives us the most acceptable sort times across browsers.

neonstalwart:
FYI my profiling includes the DOM interactions.

Matt-Esch:
yolo

PR #200, opened 2015-02-28
Special case applying xmlns attributes.

ndrwhr-inkling:
Not all attributes can be set using the dot notation like you would for `className` and `id`.  In fact most attributes should be set using setAttribute and setAttributeNS. Using the dot notation incorrectly results in attributes being dropped when rendered as actual dom elements.

In our case we are working full HTML documents that must include an xmlns attribute. Performing any parsing and then rendering of vdom elements results in missing attributes as mentioned above.

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -187,6 +187,17 @@ test("node class name is applied correctly", function (assert) {
     assert.end()
 })
 
+test("node xmlns is applied correctly", function(assert){
+    var vdom = h("html", { xmlns: "http://www.w3.org/1999/xhtml" })
+    var dom = render(vdom)
+    assert.notOk(dom.id)
+    assert.notOk(dom.className)
+    assert.equal(dom.tagName, "HTML")
+    assert.equal(dom.getAttribute("xmlns"), "http://www.w3.org/1999/xhtml")
+    assert.equal(dom.childNodes.length, 0)
+    assert.end()
+})
+
 test("mixture of node/classname applied correctly", function (assert) {
     var vdom = h("#override.very", { id: "important", className: "pretty"})
     var dom = render(vdom)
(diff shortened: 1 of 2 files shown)


Matt-Esch:
Sorry this isn't the right way to support namespaced attributes in virtual-dom.

First of all, the object you provide to virtual-hyperscript is a set of dom **properties**. They are not attributes at all. Yes, there is a difference between attributes and properties. If you wish to set an attribute instead of a property you can do this by setting a value on the attributes object

```js
h('html', { attributes: { 'xmlns': 'http://www.w3.org/1999/xhtml' } });
```

This is the most efficient way to set attributes on the node, if you really must set an attribute, but the attributes will be set without a namespace. `xmlns` is not namespaced. Generally speaking you will see namespaced attributes prefixes with a namespace identifier `<identifier>:<attrName>`

Attribute namespaces are completely orthogonal to node namespaces. so reading a namespace property from the node itself isn't sufficient. A DOM node can have attributes with different namespaces.

Currently if you want to use namespaced attributes you must use the attribute hook. For an example of heavy use of namespaced attributes, the svg function makes heavy use of these hooks to set namespaced attributes on the svg nodes.

https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/svg.js


Also, other cool use cases for `attributes` is to set `data-` attributes. So the `DataSet` interface doesn't play nicely or consistently across all browsers. The easiest way to annotate a node with `data-` attributes is through the attributes hash:

```
h('div', { attributes: { 'data-test': 'some string annotation' } });
```

This allows you to play nicely with the frameworks and tools that depend on such annotations.

ndrwhr-inkling:
@Matt-Esch my bad. Thanks for the explanation!

PR #206, opened 2015-03-17
Added link to original cmdline html2hyperscript utility

unframework:
Huge thanks to @twilson63 for making my original hacky `html2hyperscript` work as an online converter tool!

I figured I'd still add the link to the original repo (https://github.com/unframework/html2hyperscript) because it works on command line which may be helpful to some folks doing manual legacy conversion.


kuraga:
Is it like https://github.com/alexmingoia/jsx-transform ?

unframework:
@kuraga not quite, this is meant to be a one-time tool for hand-converting legacy project source code, with care taken to preserve original whitespace and even comments. Whereas `jsx-transform` is for dynamic serving or CI builds.

PR #207, opened 2015-03-18
add custom element type extensions support

rkusa:
Add support for custom elements extending existing HTML elements, e.g.:

``` html
<a is="custom-link">...</a>
```

### DIFF
--- a/vdom/create-element.js
+++ b/vdom/create-element.js
@@ -26,11 +26,20 @@ function createElement(vnode, opts) {
         return null
     }
 
-    var node = (vnode.namespace === null) ?
-        doc.createElement(vnode.tagName) :
-        doc.createElementNS(vnode.namespace, vnode.tagName)
-
     var props = vnode.properties
+    var typeExtension = props.attributes && props.attributes.is
+    var node
+
+    if (typeExtension) {
+        node = (vnode.namespace === null) ?
+                doc.createElement(vnode.tagName, typeExtension) :
+                doc.createElementNS(vnode.namespace, vnode.tagName, typeExtension)
+    } else {
+        node = (vnode.namespace === null) ?
+                doc.createElement(vnode.tagName) :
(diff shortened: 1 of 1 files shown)


neonstalwart:
got a test?

neonstalwart:
the reason i ask about the test is that i believe things will break if you don't have a value for `typeExtension`.  see https://github.com/lhorie/mithril.js/blob/b0f85a71319d9c67ef03786c1c3031e001137070/mithril.js#L258-L259

rkusa:
As of the [spec](http://w3c.github.io/webcomponents/spec/custom/):
> Let TYPE be typeExtension, or localName if typeExtension is not present

I did not add a unit test, because it would probably fail on older browsers.
A test snippet (put into the root directory of virtual-dom):

```html
<!DOCTYPE html>
<html>
<head>
  <title>test</title>
  <script type="text/javascript" src="dist/virtual-dom.js"></script>
</head>
<body>
  <script type="text/javascript">
    var proto = Object.create(HTMLAnchorElement.prototype)
    proto.attachedCallback = function() {
      alert('WORKS')
    }

    var CustomAnchorElement = document.registerElement('custom-link', {
        prototype: proto,
        extends: 'a'
    })

    var h = virtualDom.h
    var tree = h('div', {}, [
      h('a', { attributes: { is: 'custom-link' } }, [String('foo')])
    ])

    document.body.appendChild(virtualDom.create(tree))
  </script>
</body>
</html>
```

Thanks for having a look!

rkusa:
Just saw that I get 

```html
<a is="A">...</a>
```

in some browsers. I've updated the PR

neonstalwart:
the code looks good.  i just wish there was some way to write a test for this.

@Matt-Esch any thoughts on how to add a test for this?

rkusa:
Anything I can do to help getting this merged? Thanks!

PR #212, opened 2015-04-02
Allows hooks to also receive a reference of the vpatch instance.

jails:
This PR allows hooks to also receive a reference of the vpatch instance.
When a node is created, there was no easy way to retreive all the properties used for creating the node.
This would be particularly useful when the vdom is automatically generated from some html template. This way, the hook system can also be used in a wider range.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -4,10 +4,19 @@ Hooks are functions that execute after turning a VNode into an Element. They are
 Will Work
 ```javascript
 var Hook = function(){}
-Hook.prototype.hook = function(node, propertyName, previousValue) { 
+Hook.prototype.hook = function(node, propertyName, previousValue, vPatch) {
   console.log("Hello, World")
 }
-createElement(h('div', { "my-hook": new Hook() }))
+Hook.prototype.unhook = function(node, propertyName, nextValue, vPatch) {
+  console.log("Bye, World")
+}
+var first = h('div', { "my-hook": new Hook() })
+var second = h('div')
+
+var elem = createElement(h('div', first))
+var patches = diff(first, second)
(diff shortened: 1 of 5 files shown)


Zolmeister:
I am worried that exposing the `vpatch` directly opens up the API surface too much.
Might there be an alternative solution in user-land? (or at least define a new data structure for containing the v-node construction)

jails:
Maybe hooks was not the concept I was looking for. The documentation stated that hooks are "mechanisms for executing functions after a new node has been created". But looks likes they are more "mechanisms to control how a node attribute should behave on a change".

Did I miss something or there's not really a mechanism which allow to execute a function after a new node has been created ?

jails:
Closed in flavor of #213

littleloops:
@jails: Hooks aren't specifically for controlling node attributes. They are for doing things immediately after a virtual element has been turned into a real element (presumably things you couldn't do while the element was still virtual). 

For example, say you want an element to focus on the next redraw. There is no property you can set on the `HTMLElement` to make that happen, because you have to call the `HTMLElement.focus()` method. So you create a hook to do it instead (although waiting for a new animation frame / tick, because hooks are called while the rendering is still happening).

Although this might not be the problem for it (it looks like you've already moved on), I personally don't see any issue with the vpatch being passed to hooks, but then I don't mind virtual-dom giving me enough rope to hang myself if it lets me tie a few new knots :)

jails:
Oh ok thanks the head up @littleloops! Yeah I finally moved on because if Hooks gave some control over created node, I realized that beeing able to control the rendering step (i.e overriding "vdom/create-element.js" #213) was definilty the logic I needed to be able to customize.

PR #213, opened 2015-04-03
Allows to customize the rendering function.

jails:
After some investigation, I created a custom "vdom/create-element.js" function to catch the node creation step and to be able to perform some custom tasks. However I realized that there was a hard coded reference of the default "vdom/create-element.js" rendering function in "vdom/patch-op.js" and I didn't find a way to inject my custom rendering function.

In this PR there is a minor change which make the rendering function customizable the same way the patch function is.

Is it something acceptable ?

### DIFF
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -3,7 +3,6 @@ var applyProperties = require("./apply-properties")
 var isWidget = require("../vnode/is-widget.js")
 var VPatch = require("../vnode/vpatch.js")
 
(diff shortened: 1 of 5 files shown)


littleloops:
:+1:

Matt-Esch:
I'd love to merge this, I think it's the right way to go, but not without a test ;)

jails:
No big deal, just squashed the commits.
I think the tests now cover all differents use cases.

PR #215, opened 2015-04-04
Use native run-browser

kuraga:


Matt-Esch:
Why not update to latest if we're going to update it at all?

kuraga:
There are errors, then. Maybe API has been changed.
My opinion: set an equivalent version. And update when we'll update others dependencies.
(Note: `2.0.1` is strongly equiv to the current reference.)

PR #216, opened 2015-04-04
Remove old links [ci skip]

kuraga:


Matt-Esch:
Some of these are probably going to come back, but it's ok to remove them for now.

PR #217, opened 2015-04-04
Support NodeJS 0.12

kuraga:


kuraga:
`io.js` fails: https://travis-ci.org/Matt-Esch/virtual-dom/jobs/57133430
Any ideas about it?

kuraga:
`io.js` fails: https://travis-ci.org/Matt-Esch/virtual-dom/jobs/57133430
Any ideas about it?

Matt-Esch:
This is probably a bug in tap-spec (I've seen the same thing on windows and mac, but never linux). It's something to do with (a)sync writes with console.log vs fs.write(Sync).

Would be good to update tap-spec to latest to see if the bug has been fixed, if not I'd prefer a pr on tap-spec over removing it.

jails:
Should be solved in #225

Matt-Esch:
merged #225

PR #221, opened 2015-04-04
Minor test refactoring.

jails:
The fact that in the modified test hook1 & hook2 was an empty object it was possible to change hook1 references to hook2 in asserts and vice versa w/o breaking any assert (i.e assets here are based on deepEqual). Also the fact there was a common stack of calls log, it wasn't possible to check specificaly which hook has been called with the expected call. This PR is just a minor refactoring to make the test a bit more robust.


Matt-Esch:
This seems much cleaner

PR #224, opened 2015-04-06
Fixes a circular reference infinite loop when runned with `run-browser.`

jails:
I don't now if it's related to my config but looks like `'assertEqualDom()`' run into an infinite loop when trying to run tests locally with:
```
run-browser test/index.js -b | tap-spec
```

Does anybody experienced the same issue ?

### DIFF
--- a/test/lib/assert-equal-dom.js
+++ b/test/lib/assert-equal-dom.js
@@ -22,6 +22,8 @@ function areEqual(a, b) {
             key !== "self" &&
             key !== "outerHTML" &&
             key !== "innerHTML" &&
+            key !== "spellcheck" &&
+            key !== "bind" &&
             "" + parseInt(key, 10) !== key
         ) {
             if (key === "ownerDocument") {


test/lib/assert-equal-dom.js
```diff
@@ -31,7 +32,12 @@ function areEqual(a, b) {
                 return equalStyle(a[key], b[key])
             }
             if (typeof a === "object" || typeof a === "function") {
-                if (!areEqual(a[key], b[key])) {
+                if (cache.indexOf(a) !== -1) {
```
neonstalwart: i forget which is the lowest version of IE @Matt-Esch wants to support but IE9 is the earliest to support `Array.prototype.indexOf` and i think the plan is to support earlier than that.

Matt-Esch: I want to support IE8 at least because I know people out there in the real world still have to suffer through it. It's not like we are sacrificing features.

Basically, there is some property here that needs to be excluded from assertEqualDom. This should be fixed by finding which key contains the circular reference and ignoring it (shouldn't be difficult to find, just add some log statements and see what gets printed infinitely).

staltz: FWIW not even the IE team wants to support IE8 anymore. https://github.com/emberjs/rfcs/pull/45#issuecomment-89627765

jails:
Ok, thanks for the head up guys. Well I'm not sure I understood why this properties make the tests to hang but after filtering them out, I'm now able to run tests using `run-browser` locally.

jails:
closed in flavor of #225

PR #229, opened 2015-04-13
Allow minimal `vNode`

tomekwi:
Let’s make the API more forgiving to other implementations :)

**Why:** I’m working on a very light, partial implementation of `virtual-dom`. I want to use it in a browser-facing library, so I can’t afford bundling 20-odd kB minified code. But I want to keep my lib compatible with `virtual-dom` to make it extensible and plugin-friendly – `virtual-dom` seems to be becoming a standard.

### DIFF
--- a/vdom/create-element.js
+++ b/vdom/create-element.js
@@ -26,6 +26,7 @@ function createElement(vnode, opts) {
         return null
     }
 
+    var tagName = vnode.tagName || "span"
     var node = (vnode.namespace === null) ?
         doc.createElement(vnode.tagName) :
         doc.createElementNS(vnode.namespace, vnode.tagName)
@@ -33,7 +34,7 @@ function createElement(vnode, opts) {
     var props = vnode.properties
     applyProperties(node, props)
 
-    var children = vnode.children
+    var children = vnode.children || []
 
     for (var i = 0; i < children.length; i++) {
         var childNode = createElement(children[i], opts)
--- a/vdom/create-element.js
+++ b/vdom/create-element.js
@@ -26,7 +26,7 @@ function createElement(vnode, opts) {
         return null
     }
 
-    var tagName = vnode.tagName || "span"
+    var tagName = vnode.tagName || "div"
     var node = (vnode.namespace === null) ?
         doc.createElement(vnode.tagName) :
         doc.createElementNS(vnode.namespace, vnode.tagName)


tomekwi:
I can write a test or two if you guys like the idea.

vdom/create-element.js
```diff
@@ -26,6 +26,7 @@ function createElement(vnode, opts) {
         return null
     }
 
+    var tagName = vnode.tagName || "span"
```
staltz: Any particular reason why the default tagName is "span" and not "div"?

tomekwi: A span is the most minimal HTML element. Until recently it [directly implemented the `HTMLElement` interface](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span). HTML5 has changed this to [`HTMLSpanElement`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSpanElement), which is really nothing but an alias.

But it can just as well be a `div`, since it seems to match the conventions of `virtual-dom` better.

staltz: How does span's minimality affect your library? span is inline and div is block, I'd guess div is the most neutral of all.

tomekwi: It has no relevance. At the moment I’m just checking out what I need to implement to make my lib compatible with `virtual-dom`. So let’s make it a `div` :)

Raynos:
:-1: 

Don't default values; just define all the fields of the virtual node yourself.

staltz:
Yeah I agree with @Raynos . The place for defaults is virtual-hyperscript or an alternative to virtual-hyperscript

kuraga:
`tag` should not be defaultable, but why not for `children`?

staltz:
Good question, since one line later we use `children.length` in the for loop

Raynos:
All objects must be type stable. No optional fields. If you want to write slow code do it somewhere else.

tomekwi:
> No optional fields. If you want to write slow code do it somewhere else.

`<value> || <default>` doesn’t make any performance difference. The `|| <default>` doesn’t even get executed when you provide the `<value>`.

See for yourselves: http://jsbin.com/joqopi/1/edit?js,console

But I understand if it doesn’t match @Raynos’ preference. If so, feel free to close the PR.

Raynos:
The performance penalty is de-optimizing the function to be polymorphic on `vnode`. you want `vnode` to always have the same hidden class; i.e. the same map; i.e. its an object with well defined set of fields.

Zolmeister:
@Raynos agreed.
 (for those unaware: http://mrale.ph/blog/2015/01/11/whats-up-with-monomorphism.html).

tomekwi:
Wow, thanks for the link. A great read.

Fair enough. Then how about making all properties required? I saw that most of them are optional – hence the idea of making all of them like that:

```js
createElement(
  { tagName: "div"
  , children: []
  , type: "VirtualNode"
  , version: "2"
});
```

What’s your opinion on the rest then? Should they be required? I see that `h()` produces the following in addition to those I’ve listed:

- `properties`
- `key`
- `namespace`
- `count`
- `hasWidgets`
- `hasThunks`
- `hooks`
- `descendantHooks`

neonstalwart:
@tomekwi when you use the [vnode](https://github.com/Matt-Esch/virtual-dom/blob/master/vnode/vnode.js) constructor, you should be getting consistently shaped objects by providing just a few arguments.  do you find that using that constructor is a problem?

Zolmeister:
I've changed my mind actually. `createElement` doesn't really need to be performant (It's not called often), and it should be up to the developer to consider the performance implications if they choose to de-optimize the function.

tomekwi:
@neonstalwart that’s I was about to do, seeing where the discussion was heading.

In my lib I just need a DOM abstraction to pass around as messages. An abstraction which doesn’t depend on the DOM itself, so that I don’t need to pull out the heavy weapon of browser tests.

All I really need is the property `attributes`, to update a single element. I could just as well make it a one-off solution. But I want it to be compatible with `virtual-dom` for three reasons:
- easier testing with `h()`, `html2hyperscript` and such
- pluggability – easier to write a plugin for the lib, easier to plug the lib into other projects
- scalability – say I need more than one element in one or two years

It just seemed overkill to pull in the whole constructor when a one-liner like `var vNode = {attributes, type: "VitrualNode", version: "2"}` would do.

@Zolmeister are there other functions which rely on a stable object type? TBH I don’t have any battlefield experience with `virtual-dom` – I come from the world of mithril.

Zolmeister:
@tomekwi Not sure, it's a @Matt-Esch question

Matt-Esch:
@tomekwi If what you want is a minimal DOM abstraction to play like the DOM we are actually using something like this in virtual-dom called min-document. The idea is to be the minimal viable implementation of a DOM and we kinda extend it for major use cases and refuse minor edge cases. https://github.com/Raynos/min-document

tomekwi:
@Matt-Esch thanks, it’s very interesting. I’ve been using *jsdom* for any test involving the DOM until now. *min-document* feels a lot more lightweight.

But for the library itself I need another type of abstraction. Not a new dummy DOM element – but info, what attributes to update in an existing DOM element. Just a single element, just updating attributes. I want it to be an immutable, lightweight object – I want to pass it around in RxJS-like event streams.

Here’s what I have so far: [source](https://github.com/tomekwi/drawing-board/tree/2490103/source/scripts/tools), [spec](https://github.com/tomekwi/drawing-board/tree/2490103/test). Perhaps I’ll also release it as a separate lib.

jails:
@Raynos well I'm not sure to get the performance issue you are talking about. Since any object member is modified, I don't see any reason why this PR would "break" a hidden class. And related to this PR, since the children attribute has already been defaulted here `https://github.com/tomekwi/virtual-dom/blob/patch-1/vnode/vnode.js#L15` in the VNode constructor, I'm sure to get the point of doing it in `createElement()` too. And for the tagname having `<div>` instead of `<span>` by default doesn't really matter for me, whatever is fine.

tomekwi:
@Matt-Esch @Raynos @neonstalwart what I really meant by this PR was to stabilize the minimal `VirtualNode` signature accepted by any *virtual-dom* functions.

As @Zolmeister and @jails pointed out I don’t see anything wrong in defaulting some values, but for me they can just as well be required.

So please write down which way *virtual-dom* is heading. If you agree with the original intention, I’ll be happy to write up some tests and docs, and help make the behavior consistent across functions. Otherwise feel free to close this PR.

tomek-he-him:
Closing to remove clutter 🙂

Feel free to reopen if you think this is mergeable.

PR #231, opened 2015-04-16
Remove redundant code branch in diffProps

mtyaka:
I guess that the original idea was to put a `continue` statement
to the bottom of that `if` branch, as in:

```javascript
if (!(aKey in b)) {
    diff = diff || {}
    diff[aKey] = undefined
    continue
}
```

While that would be more performant in the case where the `aKey` property
is missing from `b`, it is slower in the more common case where the `aKey`
property is still present in `b`, so I think it makes sense to just remove
the `if` branch altogether.

I created a jsperf case comparing using `continue` with removing the `if` branch for when:
* all properties of `a` are removed from `b`: http://jsperf.com/virtual-dom-diffprops-with-emptied-props
* all propetires of `a` remain identical in `b`: http://jsperf.com/virtual-dom-diffprops-with-identical-props

### DIFF
--- a/vtree/diff-props.js
+++ b/vtree/diff-props.js
@@ -7,11 +7,6 @@ function diffProps(a, b) {
     var diff
 
     for (var aKey in a) {
-        if (!(aKey in b)) {
-            diff = diff || {}
-            diff[aKey] = undefined
-        }
-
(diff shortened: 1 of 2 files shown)


kuraga:
+1

Issue #230, opened 2015-04-15
add documentation for #213


neonstalwart:
add documentation for #213

cc @Matt-Esch @jails

kuraga:
:+1:

littleloops:
I'll write this today.

PR #236, opened 2015-04-28
rudimentary docs for custom patch/render functions [ci skip]

littleloops:
This is to fulfill #230 

I was going to write some examples, but I realized that it's actually patchRecursive being replaced by `renderOptions.patch`, rather than simply `patch-op/applyPatch`, so any example would have to have a lot of extra stuff in it (`patch/patchIndices`).

#213 was a step in the right direction, but I worry a bit that to take advantage of it there is going to be a lot of boilerplate code in there because `patch/patchIndices`, `patch/applyPatch`, and `patch/patchRecursive` are all private, so will have to be rewritten entirely if you decide you want to customize the patch function. It seems to me that the common use case would be to override `patch-op/applyPatch`, but there might be something I'm not thinking of.

@neonstalwart @jails


kuraga:
+1

PR #238, opened 2015-05-01
Removing noProperties and noChildren fixed variables in vnode.js

ismail-codar:
In vnode.js properties and children must be create allways new instance for all of nodes. Using noProperties and noChildren variables may cause instance problems.

### DIFF
--- a/vnode/vnode.js
+++ b/vnode/vnode.js
@@ -6,13 +6,10 @@ var isVHook = require("./is-vhook")
 
 module.exports = VirtualNode
 
-var noProperties = {}
-var noChildren = []
-
 function VirtualNode(tagName, properties, children, key, namespace) {
     this.tagName = tagName
-    this.properties = properties || noProperties
-    this.children = children || noChildren
+    this.properties = properties || {}
+    this.children = children || []
     this.key = key != null ? String(key) : undefined
     this.namespace = (typeof namespace === "string") ? namespace : null
 


Zolmeister:
+1

neonstalwart:
the assumption is that vnodes are immutable.  once you create them, they don't change - i.e. *you* don't change them.  if they don't change then you don't get issues with the way the code is currently.

Zolmeister:
@neonstalwart for [Zorium](https://github.com/Zorium/zorium) we offer [syntactic sugar for hooks](https://github.com/Zorium/zorium#lifecycle-hooks).
This means modifying the vnode after it's created (https://github.com/Zorium/zorium/blob/master/src/z.coffee#L73)
How would you propose to do this without modifying the vnode directly?

ismail-codar:
@neonstalwart can be immutability problem too.
When I trying like properties.attributes.customid = v_customid++; I see some nodes have same customid value in my app. 
I can simulate problem on the codepen. But it can be a bit tricky. Because my codes hardly spesific for my app. 
As I see no advantage for using noProperties and noChildren variables but disadvantage may cause instance problems.

neonstalwart:
> How would you propose to do this without modifying the vnode directly?

@Zolmeister you have a few options... 
* provide your own custom `h` that could wrap the existing `h` (or not) and will either:
  * adjust `properties` like what is done in https://github.com/Matt-Esch/virtual-dom/blob/88534bbd34c5befbd1689fc71e2373c9145a9807/virtual-hyperscript/index.js#L44-L51
  * or `return new VNode(tag, props || {}, childNodes || [], key, namespace);`
* create a copy of `tree.properties` before you mutate it which might also be a good practice so that you don't modify the `properties` object passed by a user and they can reuse those `properties`.
```js
tree.properties = Object.create(tree.properties)
tree.properties['zorium-hook'] = child._zorium_hook
```

---

> As I see no advantage for using noProperties and noChildren variables but disadvantage may cause instance problems.

@ismail-codar there actually are advantages. virtual-dom creates a lot of objects that can potentially get discarded from one render to the next.  garbage collection of these discarded objects can hurt performance.  by reusing `noProperties` and `noChildren` it is an attempt to reduce the amount of garbage collection.

Zolmeister:
@neonstalwart Thanks, great insight.

PR #241, opened 2015-05-04
Add PropertyHook

kuraga:
...in addition to `AttributeHook`


virtual-hyperscript/test/property-hook.js
```diff
@@ -0,0 +1,87 @@
...
+    assert.equal(elem.myprop, 'second value')
+
+    patches = diff(third, fourth)
+    patch(elem, patches)
+    assert.equal(elem.myprop, undefined)
+
+    assert.end()
+})
+
+test("sets the property if previous value was not an AttributeHook", function (assert) {
```
neonstalwart: s/AttributeHook/PropertyHook/

virtual-hyperscript/test/property-hook.js
```diff
@@ -0,0 +1,87 @@
...
+    assert.equal(elem.myprop, blankAttributeNS())
+
+    patches = diff(first, second)
+    patch(elem, patches)
+    assert.equal(elem.myprop, 'the value')
+
+    assert.end()
+})
+
+test("removes the property if next value is not an AttributeHook", function (assert) {
```
neonstalwart: s/AttributeHook/PropertyHook/

neonstalwart:
why do we need this?  ...and why not just update SoftSetHook since this is what it already does?

kuraga:
In addition to `AttributeHook`. The same but for properties.

neonstalwart:
and why not just use (or open a PR to update and maybe rename) https://github.com/Matt-Esch/virtual-dom/blob/88534bbd34c5befbd1689fc71e2373c9145a9807/virtual-hyperscript/hooks/soft-set-hook.js?

kuraga:
@neonstalwart Wow! It's really the same! I saw the name "`setSoft`" and even didn't thought about the same functionality!!!

kuraga:
Sure we should save one only. But `PropertyHook` (as same as  `AttributeHook`) checks `prev && prev.type === 'PropertyHook'`. But `SoftSetHook` doesn't.

neonstalwart:
> (or open a PR to update and maybe rename)

kuraga:
Yeah, I'm talking about:
1) update or check is needless? (I think - update),
2) rename?

neonstalwart:
yeah, i think we should improve `SoftSetHook`.  

if you think `PropertyHook` is a better name for it, then leave the code in virtual-hyperscript/hooks/property-hook.js and update virtual-hyperscript/hooks/soft-set-hook.js to be something like:

```js
// DEPRECATED: use property-hook instead
module.exports = require('./property-hook');
```

kuraga:
Thanks. Closed in favor of #242

PR #243, opened 2015-05-06
Improve thunk docs

kuraga:


kuraga:
Another case I see that special case of `previous` isn't mentioned. Is it `null` or `undefined` in this case?

littleloops:
LGTM

PR #247, opened 2015-05-16
Add type properties to hooks

tcoats:
`AttributeHook` has a `type` property available on it's prototype but the other hooks don't. Providing the `type` property helps other libraries parse vdom, such as [this discussion in vdom-to-html](https://github.com/nthtran/vdom-to-html/issues/13). At the moment there is no easy way to convert a hook to html. Using `instanceof` requires the same npm package which isn't always possible.

This PR adds a type property to `ev-hook`, `focus-hook` and `soft-set-hook`.

If it makes sense it might also be nice to add corresponding `is-ev-hook`, `is-focus-hook` and `is-soft-set-hook` files in [vnode](https://github.com/Matt-Esch/virtual-dom/tree/master/vnode).

Thoughts?

### DIFF
--- a/virtual-hyperscript/hooks/ev-hook.js
+++ b/virtual-hyperscript/hooks/ev-hook.js
@@ -25,3 +25,5 @@ EvHook.prototype.unhook = function(node, propertyName) {
 
     es[propName] = undefined;
 };
+
+EvHook.prototype.type = 'EvHook';
--- a/virtual-hyperscript/hooks/focus-hook.js
+++ b/virtual-hyperscript/hooks/focus-hook.js
@@ -18,3 +18,5 @@ MutableFocusHook.prototype.hook = function (node) {
         }
     });
 };
+
+MutableFocusHook.prototype.type = 'MutableFocusHook';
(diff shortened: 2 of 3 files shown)


kuraga:
See also #248

DeTeam:
@tcoats what do you think about having `.toString` on hook's prototype?

TimBeyer:
I think @DeTeam has the better proposal here.
That way html conversion behavior is defined by whoever defines a hook and elegantly solved by polymorphism instead of starting to switch on some kind of type identifier. 
This would solve the issue completely for the future.

tcoats:
@DeTeam I think it's great!

The only drawback is that each component will need to include a stringify library - but the freedom string gives over vdom is excellent.

DeTeam:
> The only drawback is that each component will need to include a stringify library

@tcoats, not really :) it's all about being able to represent Hook as a string. vdom-to-html lib can't guess on how to do it. If we go with just using `hook.value` it doesn't fix user hooks, while having something like `toString`, `render` as part of Hook interface will allow custom hooks to be automatically handled by vdom-to-html.

tcoats:
@DeTeam In my case I'll probably need to use a stringify lib more if `toString` is added to widgets as well, which I think is a great idea. Should `toString` be added to widgets?

`toString` is an excellent and useful addition!

PR #252, opened 2015-05-20
Allow ability to send patches has pure json

drewlesueur:
Don't rely on a node's prototype for type and version.
This allows patches from the server to be applied to the client


Zolmeister:
missing tests

nolanlawson:
For the record, I wrote a separate module to do this: https://github.com/nolanlawson/vdom-as-json

PR #257, opened 2015-05-31
Add support for non-ASCII characters in virtual-hyperscript tags

### LINKED ISSUE #249: Non-ASCII symbol blows class parsing
Here’s what happens:

```js
h('.ascii.classes').properties.className
//» 'ascii classes'

h('.classesλwith.unicode»characters').properties.className
//» 'classes unicode'
```

According to [the spec](http://dev.w3.org/csswg/css-syntax/#ident-token-diagram) any non-ascii character can be used in an identifier without escaping. Escaping CSS characters may be a complex topic – but allowing non-ascii symbols should be a breeze.

Mark-Wood:
Fixes #249

### DIFF
--- a/test/main.js
+++ b/test/main.js
@@ -54,6 +54,12 @@ test("can use class selector", function (assert) {
     assert.end()
 })
 
+test("can use non-ascii class selector", function (assert) {
+    var node = h("div.ΑΒΓΔΕΖ")
+    assertNode(assert, node, "DIV", { className: "ΑΒΓΔΕΖ" })
+    assert.end()
+})
+
 test("class selectors combine with className property", function (assert) {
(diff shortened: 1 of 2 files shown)


Matt-Esch:
I believe there's a diff somewhere that improves the perf of this in general (removes the regex and split). Merging for now, but you might want to think about back porting for the upcoming patch.

neonstalwart:
> I believe there's a diff somewhere that improves the perf of this in general (removes the regex and split).

i believe it's #180

tomekwi:
@Mark-Wood thanks! :D

tomekwi:
@Matt-Esch @neonstalwart would you be so kind to release this along with other recent patches?

PR #264, opened 2015-06-07
Expose svg on global export

bitinn:
2 good things about exposing `svg` explicitly:

1. make it clear `virtual-dom` consider it a public api.
2. make new users aware of this api.

We use `svg` extensively for our svg icon rendering, would be nice to not having to include them using full path like `virtual-dom/virtual-hyperscript/svg`.

PS: I tried to find a good place to add tests. but there doesn't seem to be a test for `index.js`, is `test/main` an ok place?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 var diff = require("./diff.js")
 var patch = require("./patch.js")
 var h = require("./h.js")
+var svg = require("./svg.js")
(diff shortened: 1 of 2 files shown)


adamyonk:
Is there something holding this up? I'm trying to use the 'virtualDom` global in a non-commonjs environment but I can't figure out where to get at `svg`.

PR #267, opened 2015-06-11
Added demonstration of instantiating dom-delegator

jessehattabaugh:
I had to go figure this out myself. It might be out of the scope of this module, but I think it's worth noting that dom-delegator is a separate package at least.


kuraga:
:+1:

Issue #262, opened 2015-06-02
SVG elements not working properly ?!


Rustem:
Hello, I have tried to use icons with virtual dom. So I have the following element:

```
module.exports = Icon

var BaseElement = require('./base-element')
var inherits = require('util').inherits
var h = require('virtual-dom/h')
var svg = require('virtual-dom/virtual-hyperscript/svg')

function Icon (  ) {
    var self = this
    BaseElement.call(this)
}

inherits(Icon, BaseElement)


Icon.prototype.render = function(iconName) {
    var className = 'icon--' + iconName
    return h('span.icon', {className: className}, [
             svg('svg', svg('use', {
                "xmlns:xlink": "http://www.w3.org/1999/xlink",
                "xlink:href": '#play'}, ''))
    ])
}
```

Matt-Esch:
Can you reproduce and add a link?

Tuch:
@Rustem, show example please

bitinn:
I am seeing similar problem as well. Our example works server-side (using `vdom-to-html`, which outputs plain html), but after patching on client-side we see problems:

If we do this:

```
var icon = h('svg.m-icon', [
  svg('use', {
    'xlink:href': 'icon.svg#name'
  })
]);
```

Then the we see the *right* result after patch, at least as far as chrome dev tool can tell:

```
<svg class="m-icon"><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/assets/icons.svg?v1.0.18#home"></use></svg>
```

But svg icon doesn't show up. if we then pick `Edit as HTML` in dev tool but doesn't change anything, the node re-render and svg icon shows up.

If we do this:

```
var icon = svg('svg.m-icon', [
  svg('use', {
    'xlink:href': 'icon.svg#name'
  })
]);
```

Then the we see the *wrong* result after patch, as `className` is not supported by `svg`:

```
// note the missing classname
<svg><use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/assets/icons.svg?v1.0.18#home"></use></svg>
```

While svg icon does show up, the missing class name means element is unstyled.

I can confirm this isn't caused by my css by using a random class name. And both example works with `vdom-to-html`.

bitinn:
After a bit testing, looks like this is the most compatible way:

```
icon = svg('svg', {
  'class': 'm-icon'
}, [
  svg('use', {
    'xlink:href': 'icon.svg#name'
  })
]);
```

[...]

PR #271, opened 2015-06-14
make class shorthand work with svg element

bitinn:
As commented here: https://github.com/Matt-Esch/virtual-dom/issues/262#issuecomment-111843921

If we don't parse tag in `svg` before handover to `h`, we won't be able to use `tag.class` shorthand because svg only support `class` property, not `className` property.

But this is kind a hack, I hope to see a better way.

### DIFF
--- a/virtual-hyperscript/svg.js
+++ b/virtual-hyperscript/svg.js
@@ -3,7 +3,7 @@
 var isArray = require('x-is-array');
 
 var h = require('./index.js');
-
+var parseTag = require('./parse-tag.js');
 
 var SVGAttributeNamespace = require('./svg-attribute-namespace');
 var attributeHook = require('./hooks/attribute-hook');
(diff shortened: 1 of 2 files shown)


eightyeight:
:+1: for this!

JamesKyburz:
This fix would make svg elements with a class property work again in [hyperx]!

[hyperx] uses [hyperscript-attribute-to-property] which replaces `class` with `className`

Then [https://github.com/Matt-Esch/virtual-dom/blob/master/virtual-hyperscript/svg.js#L35] is falsy and no `class` is added....

i.e hx`<svg class='foo' />` renders `<svg></svg>`

I can't think of a fix for [hyperx] because other implementations like `react` use className....

The following example test fails in a browser :(

```javascript
var test = require('tape')
var vdom = require('virtual-dom')
var hyperx = require('hyperx')
var svg = require('virtual-dom/virtual-hyperscript/svg')
var hx = hyperx(svg)

test('svg with class', function (t) {
  var tree = hx`<svg class="wow"><path d="M 8 0 L 0 0 0 8" class="squiggle"></path></svg>`
  t.equal(vdom.create(tree).outerHTML(), `<svg class="wow"><path d="M 8 0 L 0 0 0 8" class="squiggle"></path></svg>`)
  t.end()
})
```

[hyperx]:https://github.com/substack/hyperx
[hyperscript-attribute-to-property]:https://github.com/feross/hyperscript-attribute-to-property

OliverJAsh:
Can this get merged?

PR #272, opened 2015-06-14
Add support to virtual-hyperscript for false value in children

otaku:
An error of UnexpectedVirtualElement is being thrown when using [jsx-transform](https://github.com/alexmingoia/jsx-transform/) with the following use-case.

```<div>{x > 1 && 'You have more than one item'}</div>``` [Reference](https://facebook.github.io/react/tips/false-in-jsx.html)

It's being transformed to the following:

```h("h1", null, [ false ])```

The attached pull request would ignore any false values in addChild of virtual-hyperscript

### DIFF
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -69,7 +69,7 @@ function addChild(c, childNodes, tag, props) {
         for (var i = 0; i < c.length; i++) {
             addChild(c[i], childNodes, tag, props);
         }
-    } else if (c === null || c === undefined) {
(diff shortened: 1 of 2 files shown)


Zolmeister:
closed via https://github.com/Matt-Esch/virtual-dom/pull/154

otaku:
@Zolmeister I'm not trying to coerce the type of a boolean.

I would like ```false``` to be treated the same as ```null``` and ```undefined```.

durango:
:+1:

kuraga:
:+1:

PR #290, opened 2015-07-24
coerce numbers to strings in hyperscript

substack:
Several times I've run into errors when I accidentally use a number:

``` js
var h = require('virtual-dom/h');
var create = require('virtual-dom/create-element');
var node = h('div', [ 'clicked ', 1337, ' times' ]);
console.log(create(node).toString());
```

Currently, the above code will cause this very unfriendly error:

```
tmp/node_modules/virtual-dom/virtual-hyperscript/index.js:75
        throw UnexpectedVirtualElement({
              ^
Error: Unexpected virtual child passed to h().
Expected a VNode / Vthunk / VWidget / string but:
got:
1337.
The parent vnode is:
{
    "tagName": "DIV",
    "properties": {}
}
    at UnexpectedVirtualElement (/tmp/node_modules/virtual-dom/virtual-hyperscript/index.js:111:15)
    at addChild (/tmp/node_modules/virtual-dom/virtual-hyperscript/index.js:75:15)
    at addChild (/tmp/node_modules/virtual-dom/virtual-hyperscript/index.js:70:13)
    at h (/tmp/node_modules/virtual-dom/virtual-hyperscript/index.js:56:9)
    at Object.<anonymous> (/tmp/main.js:3:12)
    at Module._compile (module.js:431:26)
    at Object.Module._extensions..js (module.js:449:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Function.Module.runMain (module.js:472:10)
```

But with this patch, numbers are coerced to strings for text nodes:

```
<div>clicked 1337 times</div>
```

### DIFF
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -61,8 +61,10 @@ function h(tagName, properties, children) {
 }
 
 function addChild(c, childNodes, tag, props) {
-    if (typeof c === 'string' || typeof c === 'number') {
+    if (typeof c === 'string') {
         childNodes.push(new VText(c));
+    } else if (typeof c === 'number') {
+        childNodes.push(new VText(String(c)));
     } else if (isChild(c)) {
         childNodes.push(c);
     } else if (isArray(c)) {
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -61,7 +61,7 @@ function h(tagName, properties, children) {
 }
 
 function addChild(c, childNodes, tag, props) {
-    if (typeof c === 'string') {
+    if (typeof c === 'string' || typeof c === 'number') {
         childNodes.push(new VText(c));
     } else if (isChild(c)) {
         childNodes.push(c);
--- a/test/index.js
+++ b/test/index.js
@@ -6,6 +6,7 @@ require("./keys.js")
 require("./thunk.js")
 require("./style.js")
 require("./attributes")
+require("./non-string.js")
 
 require("../vdom/test/")
 require("../vtree/test/")
--- a/test/non-string.js
+++ b/test/non-string.js
@@ -1,0 +1,14 @@
+var test = require("tape")
+var h = require("../h.js")
+var diff = require("../diff.js")
+var patch = require("../patch.js")
+var createElement = require("../create-element.js")
+
+test("coerce numbers to strings in children array", function (assert) {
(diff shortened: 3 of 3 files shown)


Zolmeister:
see https://github.com/Matt-Esch/virtual-dom/pull/154

staltz:
```
Error: Unexpected virtual child passed to h().
Expected a VNode / Vthunk / VWidget / string but:
got:
1337.
```

Seems like a clear error IMO. It expects a string but got a number. `String(1337)` and all is good.

TimBeyer:
@staltz Clearly the problem is not that there's no clear error message, it's that this should not be an error.

The function is already doing a `typeof` check, and `VText` really doesn't care about what it's passed because it too simply does `String(text)`. So the oh so terrible type coercion is already happening at much lower levels of the library than `virtual-hyperscript`.

So why not just fix this annoying and inconsistent behavior?

markerikson:
Just started playing around with virtual-dom and some libraries that make use of it, and I have to agree, it's pretty unintuitive behavior.  I'm using the "vdt" library to turn JSX into a VDOM tree, and I was very confused when my first trivial sample wouldn't work.  Took me a while to figure out this was the reason.  Yes, I can do stuff like using `<div>{"" + myNumber}</div>` in my template, but it's ugly and silly.  Just auto-converting child content to strings seems like a very reasonable way to handle things.

Raynos:
lgtm. Let's do this.

virtual-hyperscript/index.js
```diff
@@ -61,7 +61,7 @@ function h(tagName, properties, children) {
 }
 
 function addChild(c, childNodes, tag, props) {
-    if (typeof c === 'string') {
+    if (typeof c === 'string' || typeof c === 'number') {
         childNodes.push(new VText(c));
```
Raynos:
```js
if (typeof c === 'string') {
    ...
} else if (typeof c === 'number') {
    childNodes.push(new VText(String(c)))
}
```

Raynos:
Let's make sure we also pass a string to the `VText` constructor. virtual-hyperscript should do the type conversion and the actual virtual nodes should be as strictly typed as possible.

Raynos:
Travis is messy; running tests locally.

Raynos:
Tests pass locally.

Raynos:
Published v2.1.0

PR #293, opened 2015-07-29
recursive patch call

mpal9000:
If `renderOptions.patch === module.exports` then `patch` is being called recursively.
For example in mercury/index.js -> `main()` call


Raynos:
Lgtm. Lets get this out.

Matt-Esch:
`renderOptions.patch` should really be unset if the default patch is to be used, but it doesn't hurt to guard against this edgecase

PR #296, opened 2015-07-29
WIP - Polymer support - DO NOT MERGE

ajnsit:
Hi,

I've setup this pull request to get feedback on whether it is possible to get polymer support, and whether this is the right way to do it.

Currently virtual-dom doesn't work with webcomponents or polymer when shadow dom polyfills are used. This is because when using polyfills, the Shadow dom nodes affect the native dom manipulation methods. Polymer offers a simple API to wrap all these native calls so that dom manipulation works both with and without shadow dom.

In this approach - I simply replaced all the direct dom manipulation with polymer api (if it is present), which seems to work with shadow dom.

**_Note that this doesn't actually import Polymer into the codebase (that would be unthinkable). It simply checks for the presence of polymer, and if present makes itself compatible.**_

The tests will definitely fail for this, so I've marked this as do not merge.

Let me know what you think.

### DIFF
--- a/dist/virtual-dom.js
+++ b/dist/virtual-dom.js
@@ -254,6 +254,8 @@ module.exports = patch
 var isObject = require("is-object")
 var isHook = require("../vnode/is-vhook.js")
 
+var Dom = require("../polymer-dom.js")
+
 module.exports = applyProperties
 
 function applyProperties(node, props, previous) {
@@ -286,7 +288,7 @@ function removeProperty(node, propName, propValue, previous) {
         if (!isHook(previousValue)) {
             if (propName === "attributes") {
                 for (var attrName in previousValue) {
-                    node.removeAttribute(attrName)
+                    Dom(node).removeAttribute(attrName)
                 }
             } else if (propName === "style") {
                 for (var i in previousValue) {
@@ -312,9 +314,9 @@ function patchObject(node, props, previous, propName, propValue) {
             var attrValue = propValue[attrName]
 
             if (attrValue === undefined) {
-                node.removeAttribute(attrName)
+                Dom(node).removeAttribute(attrName)
             } else {
-                node.setAttribute(attrName, attrValue)
+                Dom(node).setAttribute(attrName, attrValue)
             }
         }
 
@@ -359,6 +361,8 @@ var isVText = require("../vnode/is-vtext.js")
 var isWidget = require("../vnode/is-widget.js")
 var handleThunk = require("../vnode/handle-thunk.js")
 
+var Dom = require("../polymer-dom.js")
+
 module.exports = createElement
 
 function createElement(vnode, opts) {
@@ -390,7 +394,7 @@ function createElement(vnode, opts) {
     for (var i = 0; i < children.length; i++) {
         var childNode = createElement(children[i], opts)
         if (childNode) {
-            node.appendChild(childNode)
+            Dom(node).appendChild(childNode)
         }
     }
 
@@ -405,6 +409,8 @@ function createElement(vnode, opts) {
 // interest.
 
 var noChild = {}
+
+var Dom = require("../polymer-dom.js")
 
 module.exports = domIndex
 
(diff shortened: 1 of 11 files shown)


staltz:
If one needs to get Polymer elements working with virtual-dom, I'd say go cowboy style and hack virtual-dom internals to do what you did. However, we shouldn't really merge this (or anything like this) into virtual-dom. Instead, browsers should implement WCs.

ajnsit:
The problems with going cowboy style is that -

1. I can't check these modifications into git
2. Any other dependencies which depend on virtual-dom will have to be changed as well. So in my specific use case I will need to change atleast 2 other packages.
3. No one else benefits from these modifications.

What specifically are your objections to adding this support? Note that this doesn't actually import Polymer into the codebase (that would be unthinkable). It simply checks for the presence of polymer, and if present makes itself compatible.

staltz:
> What specifically are your objections to adding this support?

Yet another check during diff & patch which might slow it down. I'm just giving my humble opinion though, let's see what Matt and Jake have to say.

ajnsit:
Okay, that's a valid objection. However the performance impact should be minimal. The presence of polymer is checked only once per module. If not present, the wrapper function is defined as simply -

    var Dom = function(x) {return x;};

https://github.com/Matt-Esch/virtual-dom/pull/296/files#diff-cc5f465c9a3d786f52fe75997e7550b0R10

Zolmeister:
I like the idea, but would rather see it completely external to virtual-dom if possible. Let the client handle replacing the dom methods themselves. (maybe a virtual-dom-polymer project to make it easier)

Raynos:
This can be done fully externally by passing a polymer document into patch.

Just like we do for tests

ajnsit:
@Raynos do you have a code snippet on how to do this?

ajnsit:
A good approach would be to decouple rendering from virtual-dom to allow different plugins. Is that something people are interested in?

shashi:
I was using virtual-dom with Polymer 0.5, which used to work beautifully. For the benefit of others like me who are wondering why we need this, it's because Polymer 1.0 switched to a performant alternative they call [Shady DOM](https://www.polymer-project.org/1.0/articles/shadydom.html) (which is not a Shadow DOM polyfill but an API for walking the DOM while pretending there is Shadow DOM).

I think a great way to implement this is having a generic way to allow use of custom DOM APIs, and then having a small snippet (outside virtual-dom) which sets up Polymer interop. This patch minus `polymer-dom.js` almost gets us there.

shashi:
Note that the tests are failing because "no output is produced for 10 minutes" during the build process. I am interested to know if this passes all the tests.

shashi:
Oh I think what @Raynos means is that this can be done by passing a `Polymer.dom` object instead of a normal DOM object as the [`rootNode` argument to `patch`](https://github.com/Matt-Esch/virtual-dom/blob/ebcd2faeb44b3238c32a5e6d34a65c6025c75ec4/vdom/patch.js#L9). That is a great idea and looks like it is the general way of allowing custom DOM APIs that I was suggesting.

ajnsit:
@shashi thanks! I will try out passing a polymer doc to patch and report back.

shashi:
Turns out, `Polymer.dom(elem).children` gives an array of normal DOM nodes so `Polymer.dom(elem).children[0].children` e.g. will not do the look up via Shady DOM.

shashi:
Ok, here is a total hack to workaround this: https://gist.github.com/shashi/167eef348f268049460f

I can wrap a node in Sneaky (as in `new Sneaky(node)`) before passing it to `patch`.

shashi:
> Ok, here is a total hack to workaround this: https://gist.github.com/shashi/167eef348f268049460f

This works for traversal / insertion / deletion but since getters and setters are not proxied, this does not fully work...

shashi:
Updated gist, now Polymer works very well for me with #299. I just need to pass in `sneakyDoc` as the last (`renderOptions`) argument to both `createElement` and `patch`, and wrap the `rootNode` in `new Sneaky(rootNode)`.

One last issue remains which is that Polymer's `attached` lifecycle function is called twice in cases where a child is moved to a different location due to a `<template>...<content></content>...</template>` definition. This is a Polymer issue though.

Raynos:
Closing out due to lack of activity.

PR #297, opened 2015-07-30
Allow virtual-hyperscript to use spread notation

fmonniot:
Babel JSX builtin JSX transformation use a spread form for the children of a node:
```js
h("div", options, h("span"), h("span"))
```
while virtual-hyperscript only allow the use of a single array of child.

This PR adapt the `h` function to take either an array or a list of arguments and thus providing a native compatibility with existing JSX transformer.

### DIFF
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -16,13 +16,29 @@ var evHook = require('./hooks/ev-hook.js');
 
 module.exports = h;
 
-function h(tagName, properties, children) {
+function h(tagName, properties) {
+
+    var len = arguments.length;
+    var children;
     var childNodes = [];
     var tag, props, key, namespace;
 
-    if (!children && isChildren(properties)) {
-        children = properties;
-        props = {};
+    // Simulate spread operator
+    if (len >= 2) {
+        var from = 2;
+
+        if (isChildren(properties)) {
+            from = 1;
+            props = {};
+        }
+
+        if (len > from) {
+            children = Array(len - from);
+
+            for (var k = from; k < len; k++) {
+                children[k - from] = arguments[k];
+            }
+        }
     }
 
     props = props || properties || {};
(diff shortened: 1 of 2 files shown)


fmonniot:
Is that normal that travis is reporting a fail while all tests are ok ?

At least, on a fresh install and running `npm test` they are :)

Raynos:
:-1: unneeded complexity.

Also de-optimizes `h()`.

Zolmeister:
@Raynos I don't think it does https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments

kuraga:
:-1:

kuraga:
Ouh, now I've changed my mind!  Because [https://babeljs.io/docs/usage/jsx/] is main tool for JSX after [Deprecating JSTransform](https://facebook.github.io/react/blog/2015/06/12/deprecating-jstransform-and-react-tools.html) and **it really does so!**

I'll post a question about it in Babel conversation, but...

kuraga:
Still no?

jgaskins:
@kuraga I ran a [benchmark on this](http://jsperf.com/function-signatures) and there's a pretty wide performance gulf (24-92%) in every browser but Safari (6%) when using destructured args (via `arguments`) vs arrays.

kuraga:
@jgaskins Nice catch, I just trusted that comment :-( Reposted your comment there... Thanks!

jgaskins:
@kuraga All good. I generally trust Sebastian, too, but I honestly didn't know which way it would go and I'm on vacation with time to kill. ;-)

fmonniot:
If that's the case I will close this PR. Maybe a better way would be to do like the cycle.js project and create a helper function (https://github.com/cyclejs/cycle-dom/blob/master/src/cycle-dom.js). That way people can simplyise it for JSX and that doesn't penalize other.

kuraga:
@sebmck, can you read this issue and https://phabricator.babeljs.io/T2034#69859, please? Would great thanks!

kuraga:
@kittens - @sebmck, can you read this issue and https://phabricator.babeljs.io/T2034#69859, please? Would great thanks!

PR #309, opened 2015-09-17
Fix failing spec so that test suite passes again

garbles:
Checking for the HTML of an element seems to be environmentally dependent. `.toString()` failing on testling, while `.outerHTML` failing during coverage checks.


garbles:
@Matt-Esch @Raynos :tada:

PR #314, opened 2015-09-23
Add reference to dom2hscript in Readme

AkeemMcLennon:
Hello everyone,

I ended up creating a client side version of twilson63/html2hscript that uses the browser's built in DOM and CSS parser at https://github.com/AkeemMcLennon/dom2hscript

Although I'm still using html2hscript in the backend, the 200kb dependency on htmlparser2 was too large for use in production. However, this dependency is no longer necessary in the browser context.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ virtual-dom is heavily inspired by the inner workings of React by facebook. This
 ## Tools
 
 * [html2hscript](https://github.com/twilson63/html2hscript) - Parse HTML into hyperscript
+* [dom2hscript](https://github.com/AkeemMcLennon/dom2hscript) - Frontend library for parsing HTML into hyperscript using the browser's built in parser.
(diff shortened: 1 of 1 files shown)


Zolmeister:
How does this compare to https://github.com/bitinn/vdom-parser ?

AkeemMcLennon:
@Zolmeister Thanks for the link! To be honest, I haven't actually done much research into vdom-parser. 

Superficially, dom2hscript has a smaller footprint, but I'll be sure to compare the unit tests in order to investigate further.

AkeemMcLennon:
@Zolmeister After taking a closer look, the biggest meaningful difference is that vdom-parser outputs a VNode object and therefore has a dependency on virtual-dom itself. On the other hand, dom2hyperscript outputs a valid hyperscript string and has no external dependencies outside of testing.  Obviously you are probably using virtual-dom anyway, but that's one thing to consider. 

However, the two projects are fairly similar and I've submitted a pull request to fix a few bugs that I've discovered in the meantime.

PR #319, opened 2015-10-11
try catch attributes

chrisabrams:
For the error: `failed to setAttribute for X`.

There are scenarios (usually typo) where a bad property is added. `node.setAttribute` does not fail in a useful manner (script just stops).

I've setup a `try/catch` so that if a bad property is attempted to be added, on failure it is removed from the list.

I understand that the developer of said application should make sure that the bad property is not present, but at the very least I'd like this pull request to be a record of what is stopping execution.


Matt-Esch:
try catches are generally avoided as a performance issue and this error handling is dangerously hiding errors. if it throws you should be able to wrap the calls yourself if you're inclined to do that or add a global error listener.

PR #322, opened 2015-10-21
also sanity check that render returned non-null

jskz:


Matt-Esch:
Add a test? Exemplify where this is necessary

PR #325, opened 2015-10-24
Support custom element types via type extensions

craig0990:
- Moved `props` up to make it accessible before element instantiation
- Added a type variable, which is either the value of
  `properties.attributes.is` or `null`
- Amended `doc.createElement` to supply the second parameter for type
  extensions

### DIFF
--- a/vdom/create-element.js
+++ b/vdom/create-element.js
@@ -26,11 +26,13 @@ function createElement(vnode, opts) {
         return null
     }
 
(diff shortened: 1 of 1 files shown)


craig0990:
`npm run phantom` is failing both before and after changes made, and it also fails for me on the latest release tag, `v2.1.1`. If I can do anything about this, let me know.

Matt-Esch:
test?

PR #339, opened 2015-11-26
Let property patches be applied last (fixes #338)

### LINKED ISSUE #338: Properties depending on children don't take effect when set in current cycle
Like `scrollTop`: http://codepen.io/anon/pen/MaMEWg?editors=001

This is because property patches are applied before the children nodes have been patched. It seems though that properties like `scrollTop` depend on the childrens' state so that they can take effect.

Applying property patches after all other node patches would fix this.

pkamenarsky:
Fixes #338.


rksm:
What's the status of this? Is this going to be merged?

PR #340, opened 2015-11-29
Accept multiple arguments for children in h()

anandthakker:
This small variation allows h() to be used directly in place of React.createElement in babel-transform-react-jsx.  That babel transform seems to turn something like:

```jsx
<div>I am a {banana}</div>
```

Into 

```js
createElement('div', null, 'I am a', banana)
```

Instead of

```js
createElement('div', null, ['I am a', banana])
```

Which is what `h()` currently needs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "browser-split": "0.0.1",
     "error": "^4.3.0",
     "ev-store": "^7.0.0",
+    "flatten": "0.0.1",
     "global": "^4.3.0",
     "is-object": "^1.0.1",
     "next-tick": "^0.2.2",
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var isArray = require('x-is-array');
+var flatten = require('flatten');
 
 var VNode = require('../vnode/vnode.js');
 var VText = require('../vnode/vtext.js');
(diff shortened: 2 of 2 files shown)


staltz:
I'm confused, why is this related to virtual-dom if you are using React?

jollytoad:
I think it's to support JSX flavoured hyperscript. I've been using JSX with virtual-dom/h myself recently and created a simple wrapper function for h instead... https://gist.github.com/jollytoad/18e0b3c5cfda26d22bb4
Which deals with the rest args for children, and also converts data- and other '-' containing attributes as appropriate.

anandthakker:
@staltz: I'm not using React; as @jollytoad said, I want to use JSX _without_ React, and with hyperscript  instead.  This change allows the use of Babel's existing JSX transformer (`babel-transform-react-jsx`).

@jollytoad Thanks for sharing that wrapper.  I'm not familiar with the `dataset` and `-`-to-attributes pieces -- are those standard, or custom logic for your use case?

staltz:
Understood :+1:

jollytoad:
@anandthakker if you want to use markup like:

    <div data-something="foo" aria-whatever="blah">

You'll find this won't work with h as it stands, the wrapper handles this.

anandthakker:
@jollytoad ahh ok; am I understanding right that this is because DOM properties like `href` are equivalent to the element's attribute--so it's okay to just set them directly on the node--but other ones like `area-whatever` are not?

jollytoad:
https://github.com/Matt-Esch/virtual-dom/blob/master/docs/vnode.md explains it better than I could.

anandthakker:
:+1: thanks.

Given the attributes issue, this PR doesn't fully solve the intended problem.  Closing.

Wildhoney:
@anandthakker is this not still relevant? I am facing the same issues when transpiling `virtual-dom`'s `h` with Babel's `transform-react-jsx`. How did you come to solve this?

Issue #359, opened 2016-01-05
camel casing on style properties


substack:
camel-cased style properties are valid DOM properties, but not valid html.

```
var vdom = require('virtual-dom')
var tree = vdom.h('input', {
  style: { color: 'blue', fontSize: 24 }
})
console.log(vdom.create(tree).toString())
```

produces this output, which doesn't work when rendered as an html string:

```
<input style="color:blue;fontSize:24;" type="text" />
```

However, when rendered browser-side, these inline style properties work:

``` js
var vdom = require('virtual-dom')
var main = require('main-loop')
var loop = main({}, render, vdom)
document.querySelector('#content').appendChild(loop.target)

function render (state) {
  return vdom.h('input', { style: { color: 'blue', fontSize: 24 } })
}
```

Javey:
I think this is a bug of [min-document](https://github.com/Raynos/min-document), so I made a pull request for it https://github.com/Raynos/min-document/pull/32.

substack:
I also started out making a patch to min-document, but I don't think min-document can be made to fully support camel-cased styles without using proxies. The reason is that min-document uses an ordinary object to store `style` properties, so if you attempt to write two styles in different casings, the values do not merge into a single value like they do in the browser:

```
    test("can set camel case style properties", function(assert) {
        var elem = document.createElement("div")
        elem.style.fontSize = 24
        assert.equal(elemString(elem), "<div style=\"font-size:24;\"></div>")
        elem.style.backgroundColor = 'purple'
        assert.equal(elemString(elem),
            "<div style=\"font-size:24;background-color:purple;\"></div>")
        elem.style['font-size'] = 48
        assert.equal(elemString(elem),
            "<div style=\"background-color:purple;font-size:48\"></div>")
        assert.end()
    })
```

Raynos:
I think this would be possible with some extra loops; one to get all the cased keys and another to remove duplicates.

The only issue is order, but at least in v8 you know that if you keep the last one then that's the "last" key added since v8 is ordered in terms of assignment.

PR #361, opened 2016-01-06
camel case style properties

substack:
This patch fixes the display of camel-cased style properties showing up in the html (https://github.com/Matt-Esch/virtual-dom/issues/359).

I had to update some tests to check for properties using the dashed property names instead of camel-cased properties, but the changes are relatively minor.


Matt-Esch:
I'm not sure I understand why virtual-dom would need to change the casing of the properties. I understand that the html rendering is producing the wrong thing, but this isn't the job of virtual-dom to accommodate. virtual-dom is based purely on properties of the dom nodes in a browser agnostic fashion. min-document should be patched to generate correct html from the virtual dom node.

test/style-camel-case.js
```diff
@@ -0,0 +1,17 @@
+var test = require("tape")
+var h = require("../h.js")
+var diff = require("../diff.js")
+var patch = require("../patch.js")
+var createElement = require("../create-element.js")
+
+test("convert camel-case properties to dashes in string form", function (assert) {
```
Raynos: This is a good test :)

We should fix `.outerHTML` and `.toString()` in `min-document`.

test/nested-properties.js
```diff
@@ -25,18 +25,18 @@ test("dom node style", function (assert) {
...
     var rootNode = render(a)
     assert.equal(rootNode.style.border, style("border", "none"))
-    assert.equal(rootNode.style.className, style("className", "oops"))
+    assert.equal(rootNode.style["class-name"], style("className", "oops"))
     assert.equal(rootNode.style.display, style("display", "none"))
     var s1 = rootNode.style
     var equalNode = render(b)
     assert.equal(equalNode.style.border, style("border", "1px solid #000"))
-    assert.equal(equalNode.style.className, style("className", "oops"))
+    assert.equal(equalNode.style["class-name"], style("className", "oops"))
```
Raynos: I think this change is "bad". I don't think this change should happen at this level of the abstraction.

Raynos:
Agreed.

Doing `elem.style.className = x;` and `elem.style['class-name'] = x;` are perfectly valid DOM expression.

It's `min-document` that does the wrong thing and we should fix that instead.

substack:
closing since this issue is handled upstream in min-document

PR #362, opened 2016-01-13
Support comment node

Javey:
I want to render `VDOM` to a string in node as a template engine and I need conditional comment, so I hope `virtual-dom` to support comment node. Fortunately `min-document` supported it, therefore I made this patch for this need. Expected to be merged.


gpetrov:
@Matt-Esch Any chance to merge this? Looks pretty solid to me and also very needed.

PR #373, opened 2016-02-23
Guard against `Failed to execute 'removeChild' on 'Node'` error

prust:
Chrome throws `Failed to execute 'removeChild' on 'Node': The node to be removed is no longer a child of this node. Perhaps it was moved in a 'blur' event handler?` in some edge-cases.

### DIFF
--- a/vdom/patch-op.js
+++ b/vdom/patch-op.js
@@ -41,7 +41,12 @@ function removeNode(domNode, vNode) {
     var parentNode = domNode.parentNode
 
     if (parentNode) {
-        parentNode.removeChild(domNode)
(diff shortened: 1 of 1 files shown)


prust:
It turns out that we were running into this because we are using Handlebars templates and generating hyperscript on the fly. It's common with handlebars and mustache templating to have big chunks of HTML wrapped in if/else clauses. My suspicion is that these chunks were too similar to each-other and virtual-dom was converting one of these chunks into the other when the HTML got inserted or removed.

We switched to an approach that I'm guessing is more common with hyperscript: instead of adding and removing big chunks of the DOM, we're flagging the visibility of the elements. With this approach, the problem went away.

Issue #294, opened 2015-07-29
Accept non-string child nodes and discard null attributes


axefrog:
I'm using virtual-dom via [cycle.js/dom](https://github.com/cyclejs/cycle-dom), which uses it internally. I've noticed a couple of things which I can't imagine a good use case for leaving as-is:

### Falsey attribute values

```js
h('p', { className=null })
```

Generates:

```html
<p class="null"></p>
```

It would be good if a falsey value were to cause the attribute to be omitted altogether, seeing as, as far as I am aware, there is no useful difference between an empty attribute and a missing attribute.

### Children that are not one of: VNode / Vthunk / VWidget / string

Sometimes a value to be rendered is a number, a date, an object that has a `toString` override, etc. Rather than throwing an error, why are these not being `toString()`'d and rendered as strings? It certainly reduces the flexibility one has to throw things into the DOM and see what comes out.

axefrog:
Just saw #290, which takes care of numbers. Thoughts on my other suggestions though?

staltz:
These suggestions are primarily usability issues on virtual-hyperscript `h()`. Since `h()` targets VNode, I'd suggest making an alternative `h()` as a separate repo, or a wrapper to it, like https://www.npmjs.com/package/hyperscript-helpers. It's pretty easy to fork virtual-hyperscript, add your preferred conventions, and use it as a separate repo.

PR #376, opened 2016-03-09
stringify foreign object instead of throwing unexpected virtual element

refractalize:
I have a library called [plastiq](https://github.com/featurist/plastiq) which is starting to be used by a number of people and on a number of projects big and small. People (including myself) are often perplexed by errors like this:

```
Error: Unexpected virtual child passed to h().
Expected a VNode / Vthunk / VWidget / string but:
got:
{
    "name": "asdf"
}.
The parent vnode is:
{
    "tagName": "DIV",
    "properties": {
        "className": "haha"
    }
}
```

I think it would be more friendly and possibly more useful if the object was just turned into JSON and rendered. At least then you can see the contents of the object in the DOM, where it is in the DOM and you don't have to wonder why nothing is working before checking the dev tools console.

This patch `JSON.stringify`s the object and returns it into the vdom/dom output as a string.

Of course, I could put this into plastiq itself but thought this would yield better performance and be generally useful.

### DIFF
--- a/virtual-hyperscript/README.md
+++ b/virtual-hyperscript/README.md
@@ -63,7 +63,7 @@ If you call `h` with `h('div', { ev-click: function (ev) { } })` it
   set a property `'ev-foo'` on the DOM element.
 
 This means that `dom-delegator` will recognise the event handler
-  on that element and correctly call your handler when an a click
+  on that element and correctly call your handler when a click
   event happens.
 
 ## Installation


Zolmeister:
ref https://github.com/Matt-Esch/virtual-dom/issues/294
but from my perspective this will only hide errors

refractalize:
code is in [featurist/virtual-dom@stringify-object](https://github.com/featurist/virtual-dom/tree/stringify-object)

Issue #228, opened 2015-04-10
Input type range control has wrong value on initial render


danyx23:
When you add an input element of type range (slider control) that has min and max set to non-default values, and use the value attribute to set the initial slider position, then the initial render of virtual-dom will place the slider at a wrong position (in my tests all the way to the left or right). On the first patch application the slider jumps into the right spot, but the initial render is wrong.

See this gist which demonstrates the wrong initial position and that the first patch (after a 3 second delay) fixes it: http://requirebin.com/?gist=e1f8f94771d15d79b983

Encountered this when working with cycle-js. The discussion leading to the discovery of the bug is here: https://github.com/danyx23/cyclejsBMI/issues/1

neonstalwart:
this is a similar race condition to what happens when you try to set the value of a `<select>`.  in the case of a `<select>` if you set the `value` before an `<option>` has been added to match that `value` then the `value` will be set to the first `<option>`.  there are [ways to work around this](https://github.com/Matt-Esch/virtual-dom/issues/115#issuecomment-62406607) for a select.

the problem you're up against is that you're setting a `value` and then setting `max` and `min`.  since the `value` is outside the initial range (defined by the initial values of `max` and `min`) your `value` is not applied like you expect.  `value`, `min`, and `max` must be applied according to some order and if you declare `value` after `max` and `min` in your object then it will work:

```js
  return h('input', {
    className: "form-control",
    type: "range",
    min: 120,
    max: 200,
    value: value
  });
```

if you're not comfortable with this approach then you can use a hook to apply the `value` using something like `setTimeout` to ensure that the `value` is applied *after* `min` and `max` have been set.  incidentally, this is why your 2nd render works - `min` and `max` are set to values that will allow `value` to be set.

i don't think virtual-dom can do anything more to help this specifically but this kind of thing is why hooks are available.

danyx23:
Oh I see! It never occured to me that the order of the attributes would be significant. Knowing this I can easily work around the issue. Thank you!

ivan-kleshnin:
Wait JS objects are unordered by standard. So the problem can't be solved at all :scream: 
http://stackoverflow.com/questions/27906854/object-property-ordering-in-javascript

danyx23:
Yes you are of course right, yesterday night when I read the reply I assumed I must have missed something about the order of JS object keys being retained, but yes, it's just a lucky implementation detail that this works. This is not good.

ivan-kleshnin:
[...]

PR #377, opened 2016-03-10
Throw errors when passing invalid input values

chrisinajar:
Currently, if you pass any non-string into the `value` property of an input, when it does the hook patch the value will not `===` the current value of the input since inputs always return strings regardless of what value they were set to.

The end result is that if you pass something like a number in, cursor position is not maintained and it will constantly reset you to the end of line, editing the value of the input since `123 !== "123"`

### DIFF
--- a/virtual-hyperscript/index.js
+++ b/virtual-hyperscript/index.js
@@ -47,6 +47,16 @@ function h(tagName, properties, children) {
         props.value !== undefined &&
         !isHook(props.value)
     ) {
+        if (props.value !== null && typeof props.value !== 'string') {
+            throw UnsupportedValueType({
+                expected: 'String',
+                received: typeof props.value,
+                Vnode: {
+                    tagName: tag,
+                    properties: props
+                }
+            });
+        }
         props.value = softSetHook(props.value);
     }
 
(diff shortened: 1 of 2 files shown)


Zolmeister:
I wonder if there are other cases where this happens. If only we'd [listened](https://github.com/Matt-Esch/virtual-dom/pull/154#issuecomment-68166869)...

chrisinajar:
:-1: implicit type coercion is indeed the devil.

Throwing errors on unsupported types is better.

Silently behaving differently from the API docs is the worst.

dy:
@Zolmeister @chrisinajar I wonder why throwing an error instead of converting value to string? It is natural to have `h('input', {type: 'number', value: 100})`, isn't that?

Related to https://github.com/Matt-Esch/virtual-dom/issues/228

dy:
Also seems that is not relevant anymore, at least in chrome: it does not reset cursor on `node.value = 1234`, if `node.value === '1234'`, neither in FF. Although IE is still affected.

Zolmeister:
@dfcreative `value` is not a child, it's a prop. Props don't have to be strings (generally), this however is a  case (of perhaps many, as my first comment alluded to) where an assigned prop is coerced by the DOM.

```js
x = document.createElement('div')
x.value = 123
typeof x.value // number
y = document.createElement('input')
y.value = 123
typeof y.value // string
```

Edit: If you're suggesting we *specifically* silently coerce `<input>` values to strings, I am against it. Though this patch requires 'strings' for *all* `value` props, when it really should only apply to `<input>` type elements

PR #384, opened 2016-04-09
vDOM and 'style' h tag

sebjwallace:
vDOM is just a bootstrap object to initialize a virtual dom, and to provide a render method to update the virtual dom. This is entirely for convenience for using the virtual dom out-the-box without setup. This is just for browser use, as it requires the document object.

```
// example when virtual-dom is imported via script tag
var vDOM = new virtualDOM.vDOM()
vDOM.render(vTree)
```

The styles h tag renders a vNode with 'STYLE' tag and a child vText node with the text content of parsed styles.

```
h('style', {  '#foo':{color:'blue',weight:500}  })
```
Will render to a vText child with text:
```
#foo{color:blue;weight:500}
```
camelCasing is also available:
```
h('style', { '.bar':{backgroundColor:'grey'} })
// .bar{background-color:grey}
``` 

This is useful if a stylesheet is needed to be built dynamically within javascript. General classes, animations, css events (ie hover), even media queries can be declared using this tag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ compile
 .lock-wscript
 coverage
 node_modules
+examples/
--- a/dist/virtual-dom.js
+++ b/dist/virtual-dom.js
@@ -863,7 +863,7 @@ function h(tagName, properties, children) {
     // parse styles
     if (tag == 'STYLE' && properties){
         childNodes = parseStyles(properties);
-        properties = null
+        props = null
     }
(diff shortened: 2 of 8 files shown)


ajoslin:
Hey @sebjwallace. it looks like you should make a separate module for this. virtual-dom's core needs to stay as small as possible, and this is sugar instead of core functionality.

Issue #276, opened 2015-06-22
Accept iterable as children


OliverJAsh:
This would allow such things as `h('div', Immutable.Map(['foo', 'bar']))`, where `Immutable.Map` returns an ES6 iterable.

Matt-Esch:
No plan to add support for ES6 primitives at this stage. The code is stilll ES3 compliant, sorry ;)

theefer:
Added a suggested PR to add support https://github.com/Matt-Esch/virtual-dom/pull/400

PR #400, opened 2016-06-25
Add support for Iterable children parameters to h()

theefer:
Fixes https://github.com/Matt-Esch/virtual-dom/issues/276

The issue of generic support for Iterables in all JS libraries was [raised on Twitter](https://twitter.com/leeb/status/746733697093668864) by @leebyron and it would make it really nice to work with Immutable (among others).

Yes, it does rely on ES2015 primitives but it should fall back gracefully if used in another context.

This tentative submission adds support for Iterable lists for the `children` argument of `h` (easier than the `attributes` which require a map).

Feedback very much welcome on this!

PS: Hey @OliverJAsh :wave:


OliverJAsh:
Awesome! :-)

Issue #409, opened 2016-08-31
Duplicate keys is not reported in a clear way


yminsky:
Instead, it just leads to incorrect patches being created, which causes things to blow up farther down the line at some unspecified point in time.

It would be nice if duplicating keys would lead to an exception that explicitly flagged this problem.

Even better would be for the diffing to be done in a correct, if less efficient way, in that case.

bendrucker:
Duplicate keys as in duplicate keys in an object?

jgaskins:
Issue is not reported in a clear way ;-)

yminsky:
The problematic case is where you have two different vnodes with the same key. This causes the diffing algorithm to do the wrong thing (dropping nodes that shouldn't be dropped), silently.

jgaskins:
There's still a bit of ambiguity here. Can you show an example?

yminsky:
Sure. I have some examples I can put here in the morning. Just to be clear, though, the issue comes up specifically when you reuse the same key in multiple vnodes that have the same parent. By the key, I simply mean the argument called "key" in the constructor, here:

https://github.com/Matt-Esch/virtual-dom/blob/master/vnode/vnode.js#L12

yminsky:
Here's a concrete example.

vdom1: https://gist.github.com/yminsky/e01531a14bc96e9b0bb474cc07a43aa6
vdom2: https://gist.github.com/yminsky/3933efc5faf4eef314ecdeea5c0e89bf
patch: https://gist.github.com/yminsky/b9d6dc4e5e4ab67e8b9c2e6bfe728a14

The patch is the patch between vdom1 and vdom2, but setting the DOM to vdom1 and then applying the patch yields a different state than setting it to vdom2 directly.

I got these patches by dumping the JSON from the browser.

PR #433, opened 2017-05-05
throw better exception when duplicate keys are found

refractalize:
If VDOM has duplicate keys we see this error, which doesn't explain what went wrong:

```
Failed to execute 'insertBefore' on 'Node': parameter 1 is not of type 'Node'
```

This patch detects duplicate keys and throws this error instead:

```
duplicate vdom key: <key>, vdom: <vdom-selector>
```

Where `key` is the duplicate key, and `vdom-selector` is a `tag#id.class` representation of the node so you can find it.

### DIFF
--- a/test/keys.js
+++ b/test/keys.js
@@ -655,6 +655,17 @@ test('mixed keys move from i>0 to i<length-1', function (assert) {
     assert.end()
 })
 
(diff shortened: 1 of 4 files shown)


refractalize:
this should help with #409

PR #444, opened 2018-03-16
Fix input issues

dy:
This PR:

* Bumps deps (`npm i` givers error in windows now).
* Allows for string-like values for inputs in https://github.com/Matt-Esch/virtual-dom/pull/377.
* Invokes hooks after properties, that resolves class of issues https://github.com/Matt-Esch/virtual-dom/issues/228
* Splits `test/main.js` into `test/h.js`, `test/render.js`, `test/widget.js`, `test/vnode.js` and `test/patch.js` (that allows for better tests navigation).
* Adds `use strict` directives (to avoid mistakes & typos).
* Enables gentle eslint config and fixes all errors.
* Bumps nodejs for travis

cc @Matt-Esch @Raynos @Zolmeister 

If you prefer I can split thit PR into multiple.

I am going to create a transform converting react to virtual-dom (alike to preact), so I can devote some time fixing issues/PRs, if you have no time to run this repository you can provide contributor access to it.

There is a datavis project coming similar to [rect-vis](https://github.com/uber/react-vis), hoped to base it on virtual-dom, so looking forward to this PR not being ignored :)


vdom/apply-properties.js
```diff
@@ -24,6 +26,13 @@ function applyProperties(node, props, previous) {
             }
         }
     }
+
+    // apply hooks after properties
```
Zolmeister, line 30: Is there a test for this?

virtual-hyperscript/index.js
```diff
@@ -40,16 +40,18 @@ function h(tagName, properties, children) {
...
 
-    // fix cursor bug
+    // fix cursor bug #377
     if (tag === 'INPUT' &&
         !namespace &&
         props.hasOwnProperty('value') &&
         props.value !== undefined &&
         !isHook(props.value)
     ) {
+        if (typeof props.value === 'number' || props.value && props.value.valueOf !== Object.prototype.valueOf) props.value += ''
```
Zolmeister, line 50: Historically (https://github.com/Matt-Esch/virtual-dom/pull/154 - *just* for numbers) type coercion has been avoided.
My position is to continue strict type requirements.

dy: Ok, I just want to make sure there is understanding how "coercion is devil" works in this particular case. In #154 there is the opposite [opinion](https://github.com/Matt-Esch/virtual-dom/pull/154#issuecomment-124771390) as well.

 `h('input', {type: 'number', value: 123})` or
`h('input', {type: 'date', value: new Date()})`
is natural expectation, and forcing coercion is putting dogma over common sense, also that contradicts to [W3C priority principle](https://www.w3.org/TR/html-design-principles/#priority-of-constituencies): _users over authors ... over theoretical purity_.

I observed many times how blind following dogmas brings any project to state of stagnation and death (look at [component org](https://github.com/component) - many contributors, no life), and I would not like that to happen to _virtual-dom_.

vdom/apply-properties.js
dy: Yes, [here](https://github.com/Matt-Esch/virtual-dom/pull/444/files#diff-524f61060e1d8147246f863be0a1463aR66). May add a couple more tests if you like.

Zolmeister: It is not obvious to me how deferring hook calls solves the problem.
I was looking for a test that ran a hook and verified that it had been deferred.

virtual-hyperscript/index.js
Zolmeister: see https://github.com/Matt-Esch/virtual-dom/pull/377#issuecomment-373896581

Raynos:
Closing out due to lack of activity.

PR #446, opened 2018-04-14
vtree/diff: combine unhook and destroyWidget to use a single walk

nishanths:
I was reading this over the weekend and came across the following `TODO` in clearState, which seemed like a worthwhile thing to address.

    // TODO: Make this a single walk, not two
    unhook(vNode, patch, index)
    destroyWidgets(vNode, patch, index)

However, I have no context on whether this change will result in a measurable performance improvement. I'm happy to add benchmarks, if necessary.

### DIFF
--- a/vtree/diff.js
+++ b/vtree/diff.js
@@ -122,40 +122,6 @@ function diffChildren(a, b, patch, apply, index) {
     return apply
 }
 
-function clearState(vNode, patch, index) {
-    // TODO: Make this a single walk, not two
-    unhook(vNode, patch, index)
-    destroyWidgets(vNode, patch, index)
-}
-
-// Patch records for all destroyed widgets must be added because we need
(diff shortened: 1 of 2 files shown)


vtree/diff.js
```diff
@@ -425,3 +358,50 @@ function appendPatch(apply, patch) {
         return patch
     }
 }
+
+// clearState does the following:
+// - Add patch records for destroyed widgets. This is required because we
+//   need a DOM node reference for the destroy function.
+// - Execute hooks when two nodes are identical.
+function clearState(vNode, patch, index) {
```
nishanths, line 366: Moved this to the bottom to make the diff less messy.

nishanths:
One of the commits in this branch, 269cccb, is from #447. I can rebase this branch once #447 is merged.

